PR #11, opened 2021-08-05
Add the GetCapabilities message to the D-Bus server.

LawnGnome:
The [D-Bus notification protocol](https://specifications.freedesktop.org/notification-spec/latest/ar01s09.html) requires `GetCapabilities` to be defined; not having it means that Firefox (at least) falls back to its own, non-native notifications instead of sending notifications to swaync.

I believe the list of capabilities matches what swaync can currently handle, based on a quick and dirty set of tests, but it's certainly possible I'm wrong. The protocol spec linked above includes a list of possible capabilities.

### DIFF
--- a/src/main.vala
+++ b/src/main.vala
@@ -226,6 +226,14 @@ namespace SwayNotificatonCenter {
             notiWin.close_notification (id);
         }
 
+        public string[] GetCapabilities () throws DBusError, IOError {
+            return {
+                "actions",
+                "body",
+                "persistence",
+            };
+        }
+
         public void GetServerInformation (out string name,
                                           out string vendor,
                                           out string version,


ErikReider:
Great. Must have missed it while reading through the spec 😅. I'll take a look at it in the morning because it's 1:30am

Jaakkonen:
For me this makes Firefox and Chromium send notifications fine on first launch. Before I managed to get the notifications working fine by running mako when starting my browser, then killing it and starting `swaync`. Testing with https://www.bennish.net/web-notifications.html. :)

ErikReider:
LGTM but it doesn't work with Chromium. It at least works with Firefox. Is that a Chromium bug?

ErikReider:
@Jaakkonen I opened a new issue for the chomium bug #13

PR #30, opened 2021-12-02
Added option 'center' for 'positionX' in config

christopher-l:
This was very much trial and error, but it seems to work.

image

image


ErikReider:
LGTM :)

PR #33, opened 2021-12-19
Add multiple settings and other improvements

berbiche:
Hi,

Thanks for this notification center, I've long waited for a GTK notification center with a gnome/pantheon-like interface that worked on Sway.

This PR includes many unrelated commits that I can split-up if you want.
I recommend reviewing each commit individually.

##### Commit 1

The first commit enables showing an ellipsis when a notification's summary is too long and the body is too long.
It also enables line wrapping for overly long notifications.

image
image

The behavior can be tested with `notify-send "Ellipsis test $(python -c 'print("A"*200)')" "$(python -c 'print("A"*800)')"`.

Unfortunately, the notification window is buggy when line wrapping is enabled and does not expand to the size of the content.
The bug is not present in the notification center.

Line wrapping requires the `lines` property to be set, the maximum amount of lines could be made configurable.

##### Commit 2

The second commit makes the notification center's width and height configurable.
The option `fit-to-screen` was added, it is incompatible with `notification-center-height` since gtk-layer-shell anchors override the requested height and bind to the edges.

##### Commit 3

The third commit makes the notification center's margins configurable with GtkLayerShell.

##### Commit 4

The fourth commit adds a configurable timeout for critical notifications.

##### Commit 5

The fifth commit will make the notification center smoothly scroll to the lastest notification added.
Previously, it would jump directly with no indications.
Animations can be disabled globally in GTK with the appropriate option.

##### Commit 6

The sixth commit makes the notification window's width configurable.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -3,6 +3,6 @@
   "positionY": "top",
   "timeout": 10,
   "timeout-low": 5,
-  "keyboard_shortcuts": true,
+  "keyboard-shortcuts": true,
   "image-visibility": "always"
 }
--- a/src/config.json
+++ b/src/config.json
@@ -1,6 +1,10 @@
 {
   "positionX": "right",
   "positionY": "top",
+  "margin-top": 0,
+  "margin-bottom": 0,
+  "margin-right": 0,
+  "margin-left": 0,
   "timeout": 10,
   "timeout-low": 5,
   "keyboard-shortcuts": true,
--- a/src/config.json
+++ b/src/config.json
@@ -7,6 +7,7 @@
   "margin-left": 0,
   "timeout": 10,
   "timeout-low": 5,
+  "timeout-critical": 0,
   "keyboard-shortcuts": true,
   "image-visibility": "always"
 }
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -152,6 +152,11 @@ namespace SwayNotificatonCenter {
          */
         public int notification_center_width { get; set; default = 500; }
 
+        /**
+         * Notification window's width, in pixels.
+         */
+        public int notification_window_width { get; set; default = 500; }
+
         /* Methods */
 
         /**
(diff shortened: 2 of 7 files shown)


src/notification/notification.vala
```diff
@@ -316,10 +319,14 @@ namespace SwayNotificatonCenter {
...
-                    return;
+                    if (timeout_critical_delay == 0) {
+                        return;
+                    }
+                    timeout = timeout_critical_delay * 1000;
+                    break;
             }
             uint ms = param.expire_timeout > 0 ? param.expire_timeout : timeout;
-            if (param.expire_timeout != 0) {
+            if (ms != 0) {
```
berbiche, line 329: I'm not sure about this change. The previous check looked like a logic error to me.

ErikReider: Woops. Yeah, it should be `if (ms != 0) {`

ErikReider:
Cool! I would appreciate it if you split-up this PR :)

ErikReider:
Oh yeah! I forgot to mention that I tried to get the ellipse feature working but I wasn't able to fix the clipping unless I used a `GtkAccelLabel` for some reason...

src/configModel/configModel.vala
```diff
@@ -107,6 +107,18 @@ namespace SwayNotificatonCenter {
...
 
+        /** The timeout for notifications with CRITICAL priority */
+        private const int _timeout_critical_def = 0;
+        private int _timeout_critical = _timeout_critical_def;
+        public int timeout_critical {
+            get {
+                return _timeout_critical;
+            }
+            set {
+                _timeout_critical = value < 1 ? _timeout_critical_def : value;
```
ErikReider, line 118: Shouldn't this check if the value is smaller than 0 instead of 1?

src/config.json
```diff
@@ -7,6 +7,7 @@
   "margin-left": 0,
   "timeout": 10,
   "timeout-low": 5,
+  "timeout-critical": 0,
```
ErikReider, line 10: This also has to be documented in the README

berbiche:
Thanks for the review.
I'll split the PR and rebase my changes.

berbiche:
Closing this since it has been split in other PRs.

PR #38, opened 2021-12-23
Make notification center width and height configurable

berbiche:
#33's second commit.

The second commit makes the notification center's width and height configurable.
The option `fit-to-screen` was added, it is incompatible with `notification-center-height` since gtk-layer-shell anchors override the requested height and bind to the edges.

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -413,6 +413,38 @@ namespace SwayNotificationCenter {
         public bool script_fail_notify { get; set; default = true; }
 #endif
 
+        /** Whether to expand the notification center across both edges of the screen */
+        public bool fit_to_screen { get; set; default = true; }
+
+        /**
+         * Notification center's height, in pixels.
+         * Set `fit_to_screen` to true to ignore the height setting.
+         */
+        private const int control_center_minimum_height = 300;
+        private int _control_center_height = control_center_minimum_height;
+        public int control_center_height {
+            get {
+                return _control_center_height;
+            }
+            set {
+                _control_center_height = value > control_center_minimum_height ? value : control_center_minimum_height;
+            }
+        }
+
+        /**
+         * Notification center's width, in pixels.
+         */
+        private const int control_center_minimum_width = 300;
+        private int _control_center_width = control_center_minimum_width;
+        public int control_center_width {
+            get {
+                return _control_center_width;
+            }
+            set {
+                _control_center_width = value > control_center_minimum_width ? value : control_center_minimum_width;
+            }
+        }
+
         /* Methods */
 
         /**
--- a/src/controlCenter/controlCenter.vala
+++ b/src/controlCenter/controlCenter.vala
@@ -36,6 +36,7 @@ namespace SwayNotificationCenter {
             this.set_anchor ();
 
             viewport.size_allocate.connect (size_alloc);
+            this.map.connect (set_anchor);
 
             // Only use release for closing notifications due to Escape key
             // sometimes being passed through to unfucused application
@@ -147,39 +148,37 @@ namespace SwayNotificationCenter {
 #endif
             GtkLayerShell.set_layer (this, GtkLayerShell.Layer.TOP);
 
-            GtkLayerShell.set_margin (this, GtkLayerShell.Edge.TOP, ConfigModel.instance.control_center_margin_top);
-            GtkLayerShell.set_margin (this, GtkLayerShell.Edge.BOTTOM, ConfigModel.instance.control_center_margin_bottom);
-            GtkLayerShell.set_margin (this, GtkLayerShell.Edge.RIGHT, ConfigModel.instance.control_center_margin_right);
-            GtkLayerShell.set_margin (this, GtkLayerShell.Edge.LEFT, ConfigModel.instance.control_center_margin_left);
+            GtkLayerShell.set_margin (this,
+                                      GtkLayerShell.Edge.TOP,
+                                      ConfigModel.instance.control_center_margin_top);
+            GtkLayerShell.set_margin (this,
+                                      GtkLayerShell.Edge.BOTTOM,
+                                      ConfigModel.instance.control_center_margin_bottom);
+            GtkLayerShell.set_margin (this,
+                                      GtkLayerShell.Edge.RIGHT,
+                                      ConfigModel.instance.control_center_margin_right);
+            GtkLayerShell.set_margin (this,
+                                      GtkLayerShell.Edge.LEFT,
+                                      ConfigModel.instance.control_center_margin_left);
 
-            GtkLayerShell.set_anchor (this, GtkLayerShell.Edge.TOP, true);
-            GtkLayerShell.set_anchor (this, GtkLayerShell.Edge.BOTTOM, true);
-            switch (ConfigModel.instance.positionX) {
-                case PositionX.LEFT:
-                    GtkLayerShell.set_anchor (this,
-                                              GtkLayerShell.Edge.RIGHT,
-                                              false);
-                    GtkLayerShell.set_anchor (this,
-                                              GtkLayerShell.Edge.LEFT,
-                                              true);
-                    break;
-                case PositionX.CENTER:
-                    GtkLayerShell.set_anchor (this,
-                                              GtkLayerShell.Edge.RIGHT,
-                                              false);
-                    GtkLayerShell.set_anchor (this,
-                                              GtkLayerShell.Edge.LEFT,
-                                              false);
-                    break;
-                default:
-                    GtkLayerShell.set_anchor (this,
-                                              GtkLayerShell.Edge.LEFT,
-                                              false);
-                    GtkLayerShell.set_anchor (this,
-                                              GtkLayerShell.Edge.RIGHT,
-                                              true);
-                    break;
-            }
+            // Anchor to north/south edges as needed
+            bool anchorTop = ConfigModel.instance.positionY == PositionY.TOP;
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.TOP,
+                                      ConfigModel.instance.fit_to_screen || anchorTop);
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.BOTTOM,
+                                      ConfigModel.instance.fit_to_screen || !anchorTop);
+
+            bool anchorLeft = ConfigModel.instance.positionX == PositionX.LEFT;
+            bool anchorRight = ConfigModel.instance.positionX == PositionX.RIGHT;
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.RIGHT,
+                                      anchorRight);
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.LEFT,
+                                      anchorLeft);
+
             switch (ConfigModel.instance.positionY) {
                 case PositionY.BOTTOM:
                     list_reverse = true;
@@ -205,6 +204,11 @@ namespace SwayNotificationCenter {
                 int val = list_reverse ? 1 : -1;
                 return a.param.time > b.param.time ? val : val * -1;
             });
+
+            // Always set the size request in all events.
+            int configuredWidth = ConfigModel.instance.control_center_width;
+            int configuredHeight = ConfigModel.instance.control_center_height;
+            this.set_size_request (configuredWidth, configuredHeight);
         }
 
         private void size_alloc () {
--- a/src/notification/notification.ui
+++ b/src/notification/notification.ui
@@ -37,7 +37,6 @@
                     <property name="orientation">vertical</property>
                     <child>
                       <object class="GtkOverlay" id="overlay">
-                        <property name="width-request">500</property>
                         <property name="visible">True</property>
                         <property name="can-focus">False</property>
                         <child>
--- a/man/swaync.5.scd
+++ b/man/swaync.5.scd
@@ -87,6 +87,22 @@ config file to be able to detect config errors
 	default: true ++
 	description: Hides the control center when clicking on notification action
 
+*fit-to-screen* ++
+	type: bool ++
+	default: true ++
+	description: Whether the control center should expand vertically to fill the screen
+
+*control-center-height* ++
+	type: integer ++
+	default: 500 ++
+	description: The control centers height in pixels.
+	This setting is ignored when _fit-to-screen_ it set to "true"
+
+*control-center-width* ++
+	type: integer ++
+	default: 500 ++
+	description: The control centers width in pixels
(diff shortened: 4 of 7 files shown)


src/controlCenter/controlCenter.vala
```diff
@@ -113,6 +113,11 @@ namespace SwayNotificatonCenter {
                 return false;
             });
             this.box.add (new TopAction ("Do Not Disturb", dnd_button, false));
+
+            if (ConfigModel.instance.notification_center_height != 0) {
+                this.default_height = ConfigModel.instance.notification_center_height;
+            }
+            this.default_width = ConfigModel.instance.notification_center_width;
```
ErikReider, line 120: Needs to be set in the set_anchor method to be able to update on a config refresh

ErikReider:
Setting the height to 0 seems to shrink it by a lot. Is it possible to set a min height to the ListBox?
image

README.md
```diff
@@ -91,6 +91,9 @@ To reload the config, you'll need to run `swaync-client --reload-config`
 - `keyboard-shortcuts`: `true` or `false`. If control center should use keyboard shortcuts
 - `image-visibility`: `always`, `when-available` or `never`. Notification image visiblilty
 - `transition-time`: uint (Any positive number, 0 to disable). The notification animation duration
+- `fit-to-screen`: `true` or `false`. If the notification center should expand to fill the screen
+- `notification-center-height`: uint (Any positive number without decimals). The notification center's height in pixels, this is ignored when the `fit-to-screen` option is set to `true`
+- `notification-center-width`: uint (Any positive number without decimals). The notification center's width in pixels
```
ErikReider, line 96: Should also be changed to `control-center-height/width`

berbiche:
I couldn't figure out how to make the height of the window expand automatically to fit all the notifications to display on the screen so I removed the logic for a height of `0`.

I set a minimum width and height of 300 pixels.

src/config.json
```diff
@@ -7,5 +7,7 @@
   "notification-window-width": 500,
   "keyboard-shortcuts": true,
   "image-visibility": "always",
-  "transition-time": 200
+  "transition-time": 200,
+  "fit-to-screen": true,
+  "control-center-width": 500
```
ErikReider, line 12: Don't forget the `contro-center-heightl`. Other than that it LGTM

WTechNinja:
I may be wrong, but I think `control-center-width` and `notification-window-width` both have a minimum width of 500 from line 41 in src/notification/notification.ui where it says
`<property name="width-request">500</property>`
Could we lower that value to like 300?

ErikReider:
> I may be wrong, but I think `control-center-width` and `notification-window-width` both have a minimum width of 500 from line 41 in src/notification/notification.ui where it says `<property name="width-request">500</property>` Could we lower that value to like 300?

@WTechNinja  Oh yeah, that's true! Hmmm... Setting the notification width from this value would conflict with #42. @berbiche We should probably combine `notification-window-width` with `control-center-width` as `width`. The `control-center-height` should then be `height` and also change the notification windows [max-height](https://github.com/berbiche/SwayNotificationCenter/blob/f11da78c3118a9a1c40c4674a54fa7c2b48c45e3/src/notiWindow/notiWindow.ui#L23)

mdjose:
> I may be wrong, but I think `control-center-width` and `notification-window-width` both have a minimum width of 500 from line 41 in src/notification/notification.ui where it says `<property name="width-request">500</property>` Could we lower that value to like 300?

Even 300px could be a hassle to low resolution screens, instead, it would be nice to allow the user control it via `min-width` and `min-height`.

> @WTechNinja Oh yeah, that's true! Hmmm... Setting the notification width from this value would conflict with #42. @berbiche We should probably combine `notification-window-width` with `control-center-width` as `width`. The `control-center-height` should then be `height` and also change the notification windows [max-height](https://github.com/berbiche/SwayNotificationCenter/blob/f11da78c3118a9a1c40c4674a54fa7c2b48c45e3/src/notiWindow/notiWindow.ui#L23)

I agree with `width` and `height`, but I'm afraid of `max-width` and `max-height` as it imply the notification could autoexpand.

berbiche:
Hi, thanks for the comments and reviews, I'll get around to finish this PR next week.

src/notification/notification.ui
```diff
@@ -37,7 +37,6 @@
                     <property name="orientation">vertical</property>
                     <child>
                       <object class="GtkOverlay" id="overlay">
-                        <property name="width-request">500</property>
```
berbiche, line 40: Removing this property was sufficient for the notification to always expand to the width of the box.

berbiche:
I've updated the PR and took into account the feedback left.

- I did not merge the `notification-window-width` and `control-center-width` because I want to use a different width for each.
- The maximum height of the scrolled window (the list of notifications when the control center is not open) was not modified, it is still limited to 600 pixels.
  I think this feature should be implemented in a different PR with additional logic (e.g. display X notifications to a maximal height of Y pixels for all displayed notifications, only display notifications that can fit in the viewport (don't show half of a notification), etc.)

src/controlCenter/controlCenter.vala
```diff
@@ -152,8 +153,10 @@ namespace SwayNotificationCenter {
             GtkLayerShell.set_margin (this, GtkLayerShell.Edge.RIGHT, ConfigModel.instance.control_center_margin_right);
             GtkLayerShell.set_margin (this, GtkLayerShell.Edge.LEFT, ConfigModel.instance.control_center_margin_left);
 
-            GtkLayerShell.set_anchor (this, GtkLayerShell.Edge.TOP, true);
-            GtkLayerShell.set_anchor (this, GtkLayerShell.Edge.BOTTOM, true);
+            // Anchor to north/south edges as needed
+            GtkLayerShell.set_anchor (this, GtkLayerShell.Edge.TOP, ConfigModel.instance.fit_to_screen || ConfigModel.instance.positionY == PositionY.TOP);
+            GtkLayerShell.set_anchor (this, GtkLayerShell.Edge.BOTTOM, ConfigModel.instance.fit_to_screen || ConfigModel.instance.positionY == PositionY.BOTTOM);
```
ErikReider, line 158: Try to make it fit at a 80 char width

ErikReider (changes requested):

ErikReider:
> I think this feature should be implemented in a different PR with additional logic (e.g. display X notifications to a maximal height of Y pixels for all displayed notifications, only display notifications that can fit in the viewport (don't show half of a notification), etc.)

Possibly related to #61. Instead of cutting the notifications off, we could stack them at the bottom

src/controlCenter/controlCenter.vala
```diff
@@ -207,6 +210,22 @@ namespace SwayNotificationCenter {
...
         }
 
+        public void on_realize () {
+            // Set a minimum size for the window
+            int configuredWidth = int.max (ConfigModel.control_center_minimum_width, ConfigModel.instance.control_center_width);
+            int configuredHeight = int.max (ConfigModel.control_center_minimum_height, ConfigModel.instance.control_center_height);
+
+            // Clip the width and height to the display dimensions
+            // if they are too big
+            Gdk.Monitor monitor = get_display ().get_primary_monitor () ?? get_display ().get_monitor (0);
```
ErikReider, line 220: This only seems to get the primary monitor and not the monitor which the Window is focused on

src/controlCenter/controlCenter.vala
```diff
@@ -207,6 +210,22 @@ namespace SwayNotificationCenter {
             });
         }
 
+        public void on_realize () {
```
ErikReider, line 213: Maybe add a short comment above like:
```vala
/** Sets the Window Width and Height and clamps the values to the monitors size */
```

Also this method needs to be called in `on_visibility_change`

src/controlCenter/controlCenter.vala
```diff
@@ -207,6 +210,22 @@ namespace SwayNotificationCenter {
...
+            // Set a minimum size for the window
+            int configuredWidth = int.max (ConfigModel.control_center_minimum_width, ConfigModel.instance.control_center_width);
+            int configuredHeight = int.max (ConfigModel.control_center_minimum_height, ConfigModel.instance.control_center_height);
```
ErikReider, line 216: 80 char limit

src/configModel/configModel.vala
```diff
@@ -350,6 +350,26 @@ namespace SwayNotificationCenter {
...
 #endif
 
+        /** Whether to expand the notification center across both edges of the screen */
+        public bool fit_to_screen { get; set; default = true; }
+
+        /**
+         * Notification center's height, in pixels.
+         * `0` will not set a maximum height, allowing it to expand.
+         */
+        public const int control_center_minimum_height = 300;
```
ErikReider, line 360: These minimum values should be private

ErikReider (changes requested):

src/controlCenter/controlCenter.vala
ErikReider: Maybe we'll need a wlroots vaapi? Hopefully We'll be able to solve it without that but I'm not sure if it'll work with X11 disabled...

ErikReider: Seems like `Gdk.Monitor monitor = this.get_display ().get_monitor_at_window (this.get_window ());` works **BUT** sometime after the map event is called. I'm not sure which event to listen to but `draw` and `check_resize` kinda work but get executed many times...

src/controlCenter/controlCenter.vala
```diff
@@ -207,6 +205,53 @@ namespace SwayNotificationCenter {
...
+         * values, taking into account the monitor width and height.
+         * If the configured values are above the monitor's width/height, they
+         * will be clipped to the monitor dimensions.
+         */
+        public void on_map () {
+            // "Constants"
+            bool anchorTop = ConfigModel.instance.positionY == PositionY.TOP;
+            bool fitToScreen = ConfigModel.instance.fit_to_screen;
+            bool anchorLeft = ConfigModel.instance.positionX == PositionX.LEFT;
+            bool anchorRight = ConfigModel.instance.positionX == PositionX.RIGHT;
```
berbiche, line 219: This is messy and ugly.
Maybe I should refactor the logic to set the anchors to a distinctive function?

src/controlCenter/controlCenter.vala
```diff
@@ -147,39 +148,36 @@ namespace SwayNotificationCenter {
...
+                                      ConfigModel.instance.fit_to_screen || !anchorTop);
+
+            bool anchorLeft = ConfigModel.instance.positionX == PositionX.LEFT;
+            bool anchorRight = ConfigModel.instance.positionX == PositionX.RIGHT;
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.RIGHT,
+                                      false);
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.LEFT,
+                                      anchorRight);
```
ErikReider, line 180: Doesn't seem to work

ErikReider (changes requested):

src/controlCenter/controlCenter.vala
ErikReider: Hmmm. Clamping the width and height values seems a little out of reach atm so it would be better to just create a new PR with that enhancement and let this PR focus on setting the width and height

src/controlCenter/controlCenter.vala
berbiche: Indeed, this is a leftover.
Will fix!

berbiche: Indeed, this is a leftover.
Will fix!

src/controlCenter/controlCenter.vala
berbiche: Done.

src/controlCenter/controlCenter.vala
```diff
@@ -207,6 +205,37 @@ namespace SwayNotificationCenter {
...
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.TOP,
+                                      fitToScreen || anchorTop);
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.BOTTOM,
+                                      fitToScreen || !anchorTop);
+            // Always set the size request in all events.
+            this.set_size_request (configuredWidth, configuredHeight);
+        }
```
ErikReider, line 237: Do we even need this method anymore? Seems like removing this and running `this.set_size_request (configuredWidth, configuredHeight);` in `set_anchor()` reduces the lines by a lot. We can merge this when that's fixed :)

src/controlCenter/controlCenter.vala
```diff
@@ -147,39 +148,36 @@ namespace SwayNotificationCenter {
...
+                                      ConfigModel.instance.fit_to_screen || !anchorTop);
+
+            bool anchorLeft = ConfigModel.instance.positionX == PositionX.LEFT;
+            bool anchorRight = ConfigModel.instance.positionX == PositionX.RIGHT;
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.RIGHT,
+                                      anchorLeft);
+            GtkLayerShell.set_anchor (this,
+                                      GtkLayerShell.Edge.LEFT,
+                                      anchorRight);
```
ErikReider, line 180: Removing the `on_map` method results in the X position being flipped. We'd need to switch `anchorLeft` and `anchorRight` to fix this

ErikReider (changes requested):

README.md
```diff
@@ -128,6 +128,9 @@ To reload the config, you'll need to run `swaync-client --reload-config`
 - `control-center-margin-right`: uint (Any positive number, 0 to disable). The margin (in pixels) at the right of the notification center
 - `control-center-margin-left`: uint (Any positive number, 0 to disable). The margin (in pixels) at the left of the notification center
 - `script-fail-notify`: bool. Sends a notification with error message if a script fails
+- `fit-to-screen`: `true` or `false`. If the control center should expand to fill the screen
+- `control-center-height`: uint (Any positive number). The control center's height in pixels, this is ignored when the `fit-to-screen` option is set to `true`
+- `control-center-width`: uint (Any positive number). The control center's width in pixels
```
ErikReider, line 133: Just added a Man file for the configuration file. Make sure to remove this and add the config options there :)

src/controlCenter/controlCenter.vala
berbiche: Indeed, that was a small mistake on my end.

src/controlCenter/controlCenter.vala
berbiche: Noted. I didn't realize the code duplication at the time.

README.md
berbiche: Done!

ErikReider (approved):
LGTM! :)

PR #39, opened 2021-12-23
Configurable notification center margins with GtkLayerShell

berbiche:
#33's third commit.

The third commit makes the notification center's margins configurable with GtkLayerShell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,10 @@ To reload the config, you'll need to run `swaync-client --reload-config`
 - `notification-window-width`: uint (Any positive number). Width of the notification in pixels
 - `hide-on-clear`: bool. Hides the control center after pressing "Clear All"
 - `hide-on-action`: bool. Hides the control center when clicking on notification action
+- `control-center-margin-top`: uint (Any positive number, 0 to disable). The margin (in pixels) at the top of the notification center
+- `control-center-margin-bottom`: uint (Any positive number, 0 to disable). The margin (in pixels) at the bottom of the notification center
+- `control-center-margin-right`: uint (Any positive number, 0 to disable). The margin (in pixels) at the right of the notification center
+- `control-center-margin-left`: uint (Any positive number, 0 to disable). The margin (in pixels) at the left of the notification center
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -91,6 +91,10 @@ To reload the config, you'll need to run `swaync-client --reload-config`
 - `keyboard-shortcuts`: `true` or `false`. If control center should use keyboard shortcuts
 - `image-visibility`: `always`, `when-available` or `never`. Notification image visiblilty
 - `transition-time`: uint (Any positive number, 0 to disable). The notification animation duration
+- `notification-center-margin-top`: uint (Any positive number, 0 to disable). The margin (in pixels) at the top of the notification center
+- `notification-center-margin-bottom`: uint (Any positive number, 0 to disable). The margin (in pixels) at the bottom of the notification center
+- `notification-center-margin-right`: uint (Any positive number, 0 to disable). The margin (in pixels) at the right of the notification center
+- `notification-center-margin-left`: uint (Any positive number, 0 to disable). The margin (in pixels) at the left of the notification center
```
ErikReider, line 97: Doesn't only apply to centered control center. It would also be less confusing if it were to be called control-center-margin-right instead

berbiche:
done :heavy_check_mark:

src/configModel/configModel.vala
```diff
@@ -149,6 +149,12 @@ namespace SwayNotificatonCenter {
          */
         public int notification_window_width { get; set; default = 500; }
 
+        /** GtkLayerShell margins around the notification center */
+        public int control_center_margin_top { get; set; default = 0; }
+        public int control_center_margin_bottom { get; set; default = 0; }
+        public int control_center_margin_left { get; set; default = 0; }
+        public int control_center_margin_right { get; set; default = 0; }
```
ErikReider, line 156: Should probably check if the value is below zero. It "works" when setting a negative value but I don't think that it's a desirable feature

ErikReider:
LGTM! We can merge this once the conflicts are resolved

PR #41, opened 2021-12-23
Smooth scroll to new animation in notification center

berbiche:
#33's fifth commit.

The fifth commit will make the notification center smoothly scroll to the lastest notification added.
Previously, it would jump directly with no indications.
Animations can be disabled globally in GTK with the appropriate option.


ErikReider:
LGTM! :)

PR #42, opened 2021-12-23
Configurable notification window width

berbiche:
#33's sixth commit.

The sixth commit makes the notification window's width configurable.


ErikReider:
Found one small issue. The notification doesn't seem to expand with the width of the window.
I'll try to see if I can change it through the browser. If not I'll comment a diff :)

ErikReider:
Oh yeah! Almost forgot that the default config valid needs to be in the config.json file

PR #51, opened 2022-01-06
Fix mistake in ExecReload directive

nightly-brew:
In https://github.com/ErikReider/SwayNotificationCenter/commit/4e35be86261a381d4f0bade7098da17152a0ab2d I introduced a systemd service unit, but the syntax used in the ExecReload directive is wrong, as systemd doesn't interpret shell constructs.

The AND operator was used to execute in chain the two reload commands, one for css and one for configs, since the swaync-client can only run one at a time.

With that unit file, the second reload command is ignored (or, better put, is passed to the first swaync-client call as argument, which does nothing).
The fix consists in using ';' to separate the two calls, as per systemd docs.


nightly-brew:
Also, I just learned it's better to implement changes in separate branches, so next time I won't do it in main, sorry.

PR #53, opened 2022-01-06
Close control center when clearing all notifications

Lucaber:
After clearing all notifications we can close the control center. I can't find a reason to keep the control center open.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -7,5 +7,6 @@
   "notification-window-width": 500,
   "keyboard-shortcuts": true,
   "image-visibility": "always",
-  "transition-time": 200
+  "transition-time": 200,
+  "hide-on-clear": false
 }
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -149,6 +149,11 @@ namespace SwayNotificatonCenter {
          */
         public int notification_window_width { get; set; default = 500; }
 
+        /*
+         * Hides the control center after clearing all notifications
+         */
+        public bool hide_on_clear { get; set; default = false; }
+
         /* Methods */
 
         /**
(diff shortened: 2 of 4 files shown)


ErikReider:
This should be a configurable option

Lucaber:
now configurable via `hide-on-clear` bool in config; disabled by default

src/config.json
```diff
@@ -7,5 +7,6 @@
   "notification-window-width": 500,
   "keyboard-shortcuts": true,
   "image-visibility": "always",
-  "transition-time": 200
+  "transition-time": 200,
+  "hide-on-clear": false
```
ErikReider, line 11: This should be noted in the README

src/controlCenter/controlCenter.vala
```diff
@@ -218,6 +218,11 @@ namespace SwayNotificatonCenter {
             } catch (Error e) {
                 stderr.printf (e.message + "\n");
             }
+
+            if (ConfigModel.instance.hide_on_clear) {
+                this.set_visible (false);
+                on_visibility_change ();
```
ErikReider, line 224: To simplify this I'd just call `set_visibility` which already does this

src/configModel/configModel.vala
```diff
@@ -149,6 +149,11 @@ namespace SwayNotificatonCenter {
          */
         public int notification_window_width { get; set; default = 500; }
 
+        /*
+         * Hides the control center after clearing all notifications
+         */
+        public bool hide_on_clear { get; set; }
```
ErikReider, line 155: add `default = false;`

ErikReider:
LGTM! :)

PR #59, opened 2022-01-15
Add build for fedora and other RPM based distros.

lexa:
I've noticed that his project doesn't have pre-build fedora packages, so I made one.

This pull request adds a [rpkg](https://pagure.io/rpkg) spec teamplate which is used to generate src.rpm and then RPMs for various targets.

As example of results produced by it, there is a copr repo: https://copr.fedorainfracloud.org/coprs/lexa/SwayNotificationCenter/

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,14 @@ The package is available on the AUR:
 - [swaync](https://aur.archlinux.org/packages/swaync/)
 - [swaync-git](https://aur.archlinux.org/packages/swaync-git/)
 
+Fedora:
+An **unofficial** build is available in copr:
+
+``` zsh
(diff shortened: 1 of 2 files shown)


ErikReider:
I'm not really sure about how redhats system works but I'm guessing that it'll need to be updated every new release of swaync?

lexa:
Yes, pretty much. As far as I know there is now way to build packages fully automatically when you do new source release.

ErikReider:
In that case you'll need to add that it's an unofficial install source due to my lack of knowledge

ErikReider:
LGTM :)

PR #77, opened 2022-02-05
Be able to use JSON dictionaries in config

ErikReider:
- [x] Deserialize
- [x] Serialize
- [x] Not break when empty (empty dict and removed entirely)
- [x] Works with config reloads

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -244,13 +244,16 @@ namespace SwayNotificatonCenter {
         }
 
         /**
-         * Extracts and returns a Glib.Object from a nested JSON Object
+         * Extracts and returns a GLib.Object from a nested JSON Object
          */
         private HashTable<string, T> extract_hashtable<T>(string property_name,
                                                           Json.Node node,
                                                           out bool status) {
             status = false;
             var tmp_table = new HashTable<string, T>(str_hash, str_equal);
+
+            // Check if T is a descendant of GLib.Object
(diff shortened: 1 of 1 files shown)


ErikReider:
What do you think? @nightly-brew

nightly-brew:
Gee you're fast and more determined than me!
Lemme take a lok first, I wann know how you pulled it!

nightly-brew:
That generic there was smooth!
yep, defenitely a fantastic work, and now I know why it didn't work for me anymore...

I tried late at night to make it work, and in between changes I deleted the node return in the deserialization function. That was driving me mad, but for some reason it wasn't giving me any error .-.

ErikReider:
Aah. Yeah... I experienced that too... Any feedback before I merge? @nightly-brew

nightly-brew:
Not really, I took a quick look before shutting off the laptop but what I saw made sense.

I couldn't verify, but I have the feeling this implementation doesn't work for nested hashtables.
If that's the case, I'd say we only need to avoid using nested dicts.

ErikReider:
> I couldn't verify, but I have the feeling this implementation doesn't work for nested hashtables.
If that's the case, I'd say we only need to avoid using nested dicts.

Yeah, it only checks one level deep

nightly-brew:
Well then, no complaints nor suggestions!
Let's merge, then I'll open a MR for the sound feature later :)

PR #87, opened 2022-02-15
X11 positioning support

ErikReider:
- [x] Positioning
- [ ] Don't overlap with Polybar (if even possible?)
- [ ] Get keyboard shortcuts working
- [ ] Try to use WindowType TOPLEVEL instead of POPUP
- [x] Get background opacity working
- [ ] Update README with text that states that this is experimental
- [ ] Print to stdout when started that X11 support is experimental
- [ ] Make sure that margins work
- [ ] Make sure that all positions work
- [ ] Make sure that all height and width config options work


berbiche:
> Don't overlap with Polybar (if even possible?)

I don't know whether EWMH would allow you to position the widget below polybar automatically, but an alternative is to set a top margin equal to polybar's height.

PR #89, opened 2022-02-16
build: add fedora build script

ldelossa:
Add a built script for fedora

Signed-off-by: ldelossa <louis.delos@gmail.com>

### DIFF
--- a/contrib/fedora-container-build.sh
+++ b/contrib/fedora-container-build.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+rm -rf ./build/
+
+dnf install -y gtk-layer-shell-devel \
+    libhandy-devel \
+    vala \
+    json-glib-devel \
+    meson \
+    cmake 
+
+meson build
+ninja -C build
+meson install -C build


ErikReider:
Does this ingrate well with the default package manager? if not, updating could cause compatibility issues in the future

ldelossa:
DNF is fedora's default package manager. All packages needed to build are present in the default repository.

ErikReider:
What I meant is if swaync will show up as installed in dnf

ErikReider:
Hmm. On Arch we have this package in the AUR which integrates very well with pacman (the default package manager) so uninstalling a package like libhandy won't work without uninstalling swaync.

Does dnf support building a local rpkg? That could be a better and safer solution

ldelossa:
Ohh I see, I was under the impression swaync was statically compiled. I see your point, you'd like to link any runtime depedencies to swaync, so if one is removed, you'd be warned its used by swaync?

ldelossa:
Okay cool, not sure if dnf offers anything like this outside of just building an RPM, but I can take a look.

ErikReider:
Hmm. Just remembered that every dependency should be statically linked

ldelossa:
If that is the case, you could remove all dependencies once the binary is built, right? The script begins to make more sense then. 

Originally I wanted to build swaync in a container, which just built the binary and dumped it to the host machine. However, meson is still required to do the actual install part. Not being able to fully encapsulate all the dependencies in the container itself made me abandon that and just push the script

ErikReider:
What happens if you uninstall all of those dependencies after `meson install`? Does it even run?

ldelossa:
Not sure. If meson uses anything like pkg-config to see if libs are avail, should fail,

ErikReider:
> Not sure. If meson uses anything like pkg-config to see if libs are avail, should fail,

If swaync launches without any errors after uninstalling those dependencies, I'll merge this :)

mcepl:
Is it really good idea? The main point of all those `*.spec` files, `debian/*` directories, etc. is that they should provide only the distro-specific information. Why to maintain it upstream, when the proper SPEC file for Fedora (and yes, I could add one for openSUSE as well) will be maintained in https://src.fedoraproject.org/ anyway?

ldelossa:
I only made this as a stop gap because the copr repo trailed behind HEAD pretty far at the time. I use this for myself to build swaync on my machine. I punted it forward incase it was useful, im not too concerned if this gets merged or not. Ill close it.

Issue #101, opened 2022-03-17
`-swb`: add a class which gets added/removed as the notification centre is toggled


neon64:
In my current waybar setup I have a few buttons which show visual toggle states , i.e.: they get a slightly different background colour etc... when activated.

I added a button to toggle swaync's notification center following this example: https://github.com/ErikReider/SwayNotificationCenter#waybar-example, but as an enhancement to that, what I thought would be really cool would be if `swaync-client -swb` would also emit new lines when the notification center is opened/closed, and it would toggle a particular class so that we could style the "toggle" state with Waybar's own css file.

e.g. something like:
```
{"text": "", "alt": "notification", "tooltip": "", "class": "notification"}
// after user opens notification center
{"text": "", "alt": "notification", "tooltip": "", "class": "notification notification-center-open"}
// after user closes notification center
{"text": "", "alt": "notification", "tooltip": "", "class": "notification"}
```

Awesome project by the way!

ErikReider:
I'll look into this :)

PR #103, opened 2022-03-17
waybar integration: emit an array of classses rather than a single class

neon64:
My bad, the example output I showed in https://github.com/ErikReider/SwayNotificationCenter/issues/101 wasn't in the correct format.

Waybar doesn't like:
```"class": "foo bar"```

^ it treats this as a single class "foo bar" (which then breaks existing styles)

Instead it wants:
```"class": ["foo", "bar"]```

### DIFF
--- a/src/client.vala
+++ b/src/client.vala
@@ -57,13 +57,15 @@ private void on_subscribe_waybar (uint count, bool dnd, bool cc_open) {
     if (count > 0) {
         tooltip = @"$(count) Notification" + (count > 1 ? "s" : "");
     }
-    print ("{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": ",state, tooltip);
+
+    string _class = @"\"$(state)\"";
     if (cc_open) {
-        print("[\"%s\", \"cc-open\"]", state);
-    } else {
-        print("\"%s\"", state);
+        _class = @"[$(_class), \"cc-open\"]";
     }
-    print("}\n");
+
+    print (
+        "{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": %s}\n",
+        state, tooltip, _class);
 }
 
 public int command_line (string[] args) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/client.vala [hunk 1]
```diff
@@ -52,15 +52,18 @@ private void on_subscribe (uint count, bool dnd, bool cc_open) {
...
     string tooltip = "";
     if (count > 0) {
         tooltip = @"$(count) Notification" + (count > 1 ? "s" : "");
     }
-
-    print ("{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": \"%s\"}\n",
-           state, tooltip, state);
+    print ("{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": ",state, tooltip);
+    if (cc_open) {
+        print("[\"%s\", \"cc-open\"]", state);
```
neon64, line 62: It would be probably nicer to build up a Vala array of classes, and then convert that array to JSON in one step.
However I'm not too familiar with Vala so I did it this way with an `if-else` for now

ErikReider:
Nice catch! I made some changes so that it only prints once :)
```
diff --git a/src/client.vala b/src/client.vala
index db7569d..7dc290e 100644
--- a/src/client.vala
+++ b/src/client.vala
@@ -57,13 +57,15 @@ private void on_subscribe_waybar (uint count, bool dnd, bool cc_open) {
     if (count > 0) {
         tooltip = @"$(count) Notification" + (count > 1 ? "s" : "");
     }
-    print ("{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": ",state, tooltip);
+
+    string _class = @"\"$(state)\"";
     if (cc_open) {
-        print("[\"%s\", \"cc-open\"]", state);
-    } else {
-        print("\"%s\"", state);
+        _class = @"[\"$(state)\", \"cc-open\"]";
     }
-    print("}\n");
+
+    print (
+        "{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": %s}\n",
+        state, tooltip, _class);
 }
 
 public int command_line (string[] args) {
```

neon64:
thanks, have integrated those changes

ErikReider (approved):
LGTM! :)

PR #111, opened 2022-04-05
Don't force-build the man-pages, use option instead

freijon:
Currently the man-pages are built regardless if the option 'man-pages' is true or false. This PR basically put's the entire man-page section in an if clause that checks if the option was selected or not.


ErikReider (approved):
LGTM! Thanks! :)

PR #121, opened 2022-05-11
Do not keep transient notifications in the CC

### LINKED ISSUE #120: Feature request: Respect transient notifications
Hi,
The Control Center doesn't seem to respect the transient flag of notifications. So doing something like `notify-send -e -t 2000 "Transient notification"` still stacks the notification in the control center, when it should just show the popup and drop the notification after it disappears.

I quickly look at the code base, and it seems to be something I could easily add. Is it something in which you would be interested?

aruhier:
Fixes #120.

Notifications can indicate in their hints that they should be transient.
A transient notification is not persistent, therefore, respects this hint and do not stack transient notifications in the control center.

Edited by @ErikReider:
Example notification
`notify-send.sh "Test Notification" '<b>Transient Notification</b>' -h "int:transient:0"`

### DIFF
--- a/src/notiDaemon/notiDaemon.vala
+++ b/src/notiDaemon/notiDaemon.vala
@@ -182,8 +182,8 @@ namespace SwayNotificationCenter {
                     noti_window.add_notification (param, this);
                 }
             }
-            // Only add notification to CC if it isn't IGNORED
-            if (state != NotificationStatusEnum.IGNORED) {
+            // Only add notification to CC if it isn't IGNORED and not transient
+            if (state != NotificationStatusEnum.IGNORED && !param.transient) {
                 control_center.add_notification (param, this);
             }
 
--- a/src/notiModel/notiModel.vala
+++ b/src/notiModel/notiModel.vala
@@ -83,6 +83,7 @@ namespace SwayNotificationCenter {
         public string desktop_entry { get; set; }
         public string category { get; set; }
         public bool resident { get; set; }
+        public bool transient { get; set; }
         public UrgencyLevels urgency { get; set; }
         /** Replaces the old notification with the same value of:
          * - x-canonical-private-synchronous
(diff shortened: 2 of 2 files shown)


src/notiDaemon/notiDaemon.vala
```diff
@@ -182,8 +182,15 @@ namespace SwayNotificationCenter {
...
-            // Only add notification to CC if it isn't IGNORED
-            if (state != NotificationStatusEnum.IGNORED) {
+            // Only add notification to CC if it isn't IGNORED and not transient
+            var transient_hint = hints.get("transient");
+            var transient = (transient_hint != null &&
+                transient_hint.is_of_type (GLib.VariantType.INT32) &&
+                    transient_hint.get_int32() == 1 ||
+                transient_hint.is_of_type (GLib.VariantType.BOOLEAN) &&
+                    transient_hint.get_boolean()
+            );
```
ErikReider, line 192: This should be ran in the notiModel.vala file where all of the other hints are parsed

ErikReider (changes requested):

aruhier:
> Are there any examples of notifications that are transient like Spotify?

Not that I'm aware of. But a user might want to trigger a notification with the transient hint, for example if a laptop's charger is plugged or unplugged, with notify-send.

src/notiDaemon/notiDaemon.vala
aruhier: So should I set a `transient` attribute and check its value here?

ErikReider: > So should I set a `transient` attribute and check its value here?

Precisely! Just like all of the other hints :)

aruhier: Done,  thanks!

src/notiModel/notiModel.vala
```diff
@@ -226,6 +227,13 @@ namespace SwayNotificationCenter {
                             resident = hint_value.get_boolean ();
                         }
                         break;
+                    case "transient":
+                        if (hint_value.is_of_type (GLib.VariantType.BOOLEAN)) {
+                            transient = hint_value.get_boolean ();
+                        } else if (hint_value.is_of_type (GLib.VariantType.INT32)) {
+                            transient = hint_value.get_int32() == 1;
```
aruhier, line 234: About this part, just to give some explanations: I saw a lot of tutorials/answers on forums that are using a int32 for the transient parameter instead of a boolean, that's why I prefer covering both cases.

ErikReider, line 234: Just add a space between the `get_int32` and `()` to make the linter happy :)

ErikReider (changes requested):

ErikReider (approved):
LGTM! Thanks! :)

Issue #122, opened 2022-05-12
PKGBUILD for aarch_64 devices


ErikReider:
Would require `gtk3-mobile` instead of `gtk3`

ragreenburg:
I made the comment on the AUR post. I could also just make a PKGBUILD that is something along the lines or `swaync-mobile`?

ErikReider:
I'd prefer `arm` instead of `mobile`. What do other arm AUR packages do?

ragreenburg:
Just from a quick look, `-arm` seems a fair bit more popular. I just wasn't sure if `gtk3-mobile` is used on other ARM devices, as I'm using a [PinePhone](https://pine64.com/product/pinephone-pro-explorer-edition/), so I'm not sure if it is just mobile specific.

ErikReider:
I can't seem to find the `gtk3-mobile` package anywhere. I've search through the official packages and AUR but nothing found? Which mirrors are you using?

ragreenburg:
I think the repo is [https://p64.arikawa-hi.me](https://p64.arikawa-hi.me). So it might be specific for the PinePhone/PINE64.

ErikReider:
Then there's nothing that I can do until Arch ships a arm64 native gtk3 package

mpldr:
https://archlinuxarm.org/packages/armv7h/gtk3 It does (At least the only ARM Arch I know of does)

ErikReider:
> https://archlinuxarm.org/packages/armv7h/gtk3 It does (At least the only ARM Arch I know of does)

Oh cool! I didn't know about that site! Then the script should in theory work but I have no way of testing it. I'll come back to this when GitHub supports running arm64 docker images for testing

PR #126, opened 2022-05-17
Add arm to pkgbuild + pkgbuild fixes

mpldr:
Also fixes some other minor issues in the PKGBUILD.

Also: annotated tags should be preferred when tagging versions (you can add the changelog directly to the tag).

Fixes: #122


mpldr:
Really, please use annotated tags. Otherwise your git package versions will be somewhat messed up

PKGBUILD
```diff
@@ -2,23 +2,27 @@
...
 build() {
-  arch-meson "${_pkgfoldername}-${pkgver}" build -Dscripting=true
-  ninja -C build
+	arch-meson "${_pkgfoldername}-${pkgver}" build -Dscripting=true
+	ninja -C build
 }
 
 package() {
-  DESTDIR="$pkgdir" meson install -C build
+	DESTDIR="$pkgdir" meson install -C build
```
ErikReider, line 27: Unnecessary change to tabs. I use `shfmt -ci -s -bn -i 4` to format these pkgbuild files so it's easier to keep it like that :)

PKGBUILD-git
```diff
@@ -1,35 +1,41 @@
...
 sha256sums=('SKIP')
 
 pkgver() {
-  cd $_pkgfoldername
-  printf "$_ver.r%s.%s" "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)"
+	cd $_pkgname
+	( set -o pipefail
+		git describe --long 2>/dev/null | sed 's/\([^-]*-g\)/r\1/;s/-/./g' ||
+		printf "r%s.%s" "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)"
+	)
```
ErikReider, line 26: Why change this? It worked perfectly fine before. This change also excludes the pkgver so the resulting version would become `swaync-git r308.a9c7af2-1` instead of `swaync-git 0.6.1.r308.a9c7af2-1`

PKGBUILD
```diff
@@ -2,23 +2,27 @@
...
+pkgdesc="A simple notificaion daemon with a GTK panel for checking previous notifications like other DEs"
 _pkgfoldername=SwayNotificationCenter
 url="https://github.com/ErikReider/$_pkgfoldername"
-arch=(x86_64)
+arch=(
+	'x86_64'
+	'aarch64' # ARM v8 64-bit
+	'armv7h'  # ARM v7 hardfloat
+)
 license=(GPL)
```
ErikReider, line 13: This should also be GPL3

ErikReider (changes requested):

ErikReider:
We should also move all of the build scripts (including rpkg.spec) into a new directory called `build-scripts` to make testing these build scripts easier

PKGBUILD
mpldr: Was done for accessibility

PKGBUILD-git
mpldr: > I'll still need to update the file in the AUR repo to prompt an update though
That's not the intended way of performing updates on development packages though. These are intended to be automatically updated by on each -Syu if there have been now commits.

> Do you know if there's a way for GitHub to automatically do that?
GitHub has long abandoned the "git-way". Maybe you can solve that using a workflow which automatically commits releases as annotated tags as well.

ErikReider: > That's not the intended way of performing updates on development packages though. These are intended to be automatically updated by on each -Syu if there have been now commits.

Makes sense :)

> Maybe you can solve that using a workflow which automatically commits releases as annotated tags as well.

Do you have any recommendations? Tagging through the terminal -> creating release -> copying the generated change-log -> adding a tag annotation could be a solution. Maybe not the best and most smooth experience though :/

mpldr: Maybe I can tinker something.

mpldr: > Do you have any recommendations?

Just remembered I have this in my .zshrc:
```shell
function tag_release {
	if [ "$1" = "" ]; then
		echo "Usage: tag_release TAG"
		exit 2;
	fi
	git shortlog $(git tag -l --sort=taggerdate | tail -1)..HEAD | git tag -a "$1" -F -
}
```

This also autogenerates a changelog and puts it into an annotated tag.

mpldr: Maybe you can also drop an upvote on https://github.com/github/feedback/discussions/4924 to let the Octocats know that it would be a good idea to have tags set as intended.

ErikReider: That works perfectly! Did not know that `shortlog` existed... Works as expected :)

ErikReider (approved):
LGTM! Thanks for the help! Future releases will be annotated accordingly :)

Issue #141, opened 2022-07-06
Waybar subscription missed opportunity


carlosV2:
Hi,

I just replaced `mako` with `swaync` and, so far, I love it.

I think there is tho a missed opportunity on the waybar subscription. You see, this is the output I'm getting:
```sh
$ swaync-client -swb
{"text": "", "alt": "none", "tooltip": "", "class": "none"}
{"text": "", "alt": "notification", "tooltip": "1 Notification", "class": "notification"}
{"text": "", "alt": "notification", "tooltip": "2 Notifications", "class": "notification"}
{"text": "", "alt": "notification", "tooltip": "3 Notifications", "class": "notification"}
{"text": "", "alt": "notification", "tooltip": "4 Notifications", "class": "notification"}
```

This is really cool but I would like to have the number of notifications displayed in the waybar too and it would be as easy as including the number only in the `text` field too. Ultimately, it is up to the user to show the number or not. If the README example is followed as is, it would still not show the number as `{}` is missing in the `format` field.

I'll try to add the number there with a bash/python script although I'm pretty sure it should be pretty straight forward to return it from the `swaync-client` too. If you point me to the right files, I can attempt a PR (provided you like this option, of course).

Thanks for a great project!

carlosV2:
If anyone is interested on this, I've managed to compose a `jq` query that does exacly this:
```sh
$ swaync-client -swb | jq -c '.text=(.tooltip|match("\\d+").string//"")'
```

ErikReider:
Great suggestion! I don't think that I'll have any time to add this but here's the swaync-client file [src/client.vala](https://github.com/ErikReider/SwayNotificationCenter/blob/440ca813c78424e933977fec87d007f1c6fee0c9/src/client.vala)

PR #142, opened 2022-07-07
Issue #141: Waybar subscription missed opportunity

carlosV2:
This PR includes the number of notifications in the text field when `swaync-client` is called with waybar subscription. (Issue #141)

With this PR, the output of the command becomes like this (please, note the `text` field being non-empty now):
```
{"text": "1", "alt": "notification", "tooltip": "1 Notification", "class": "notification"}
{"text": "2", "alt": "notification", "tooltip": "2 Notifications", "class": "notification"}
{"text": "2", "alt": "notification", "tooltip": "2 Notifications", "class": ["notification", "cc-open"]}
{"text": "2", "alt": "notification", "tooltip": "2 Notifications", "class": ["notification", "cc-open"]}
{"text": "1", "alt": "notification", "tooltip": "1 Notification", "class": ["notification", "cc-open"]}
{"text": "0", "alt": "none", "tooltip": "", "class": ["none", "cc-open"]}
{"text": "0", "alt": "none", "tooltip": "", "class": ["none", "cc-open"]}
{"text": "0", "alt": "none", "tooltip": "", "class": ["none", "cc-open"]}
{"text": "0", "alt": "none", "tooltip": "", "class": "none"}
...
```

P.S.: I couldn't find tests for this feature. Are there any?

Thank you!

### DIFF
--- a/src/client.vala
+++ b/src/client.vala
@@ -87,9 +87,10 @@ private void on_subscribe_waybar (uint count, bool dnd, bool cc_open) {
         _class = "[%s, \"cc-open\"]".printf (_class);
     }
 
+    string total = "%u".printf (count);
     print (
         "{\"text\": \"%s\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": %s}\n",
-        count.to_string(), state, tooltip, _class);
+        total, state, tooltip, _class);
 }
 
 private void print_subscribe_waybar () {
--- a/src/client.vala
+++ b/src/client.vala
@@ -88,8 +88,8 @@ private void on_subscribe_waybar (uint count, bool dnd, bool cc_open) {
     }
 
     print (
-        "{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": %s}\n",
-        state, tooltip, _class);
+        "{\"text\": \"%s\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": %s}\n",
+        count.to_string(), state, tooltip, _class);
 }
 
 private void print_subscribe_waybar () {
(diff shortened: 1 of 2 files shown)


ErikReider:
I sadly haven't had any time to write any tests at all :(

ErikReider:
A update to the README would be good too :)

carlosV2:
Ok, I've added a little something to the README.

I've also tried to fix the linting issue but because I've never worked with vala before, I'm not sure I've fixed it or made it worse... let's see (to be honest, I wasn't sure what the error really was 😅).

src/client.vala
```diff
@@ -87,9 +87,10 @@ private void on_subscribe_waybar (uint count, bool dnd, bool cc_open) {
         _class = "[%s, \"cc-open\"]".printf (_class);
     }
 
+    string total = "%u".printf (count);
     print (
-        "{\"text\": \"\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": %s}\n",
-        state, tooltip, _class);
+        "{\"text\": \"%s\", \"alt\": \"%s\", \"tooltip\": \"%s\", \"class\": %s}\n",
+        total, state, tooltip, _class);
```
ErikReider, line 93: The previous code was fine, the linter was complaining about a missing space between the `to_string` and `();`

carlosV2:
Ok, done!

All you said makes sense. I didn't realise the parenthesis had an space leading them. Anyway, it should work now 😄

ErikReider (approved):
LGTM :)

carlosV2:
Out of curiosity, what is the process this needs to follow until it ends in a linux distro? Arch to be more precise

ErikReider:
Atm it's only officially supported in the AUR. I'm actually not sure about that... I'll maybe check it out if I have some spare time. It would be nice to only need to maintain the swaync-git PKGBUILD instead of maintaining both (if that's even how it works...). I'll open an issue :)

carlosV2:
Ah ok, cool. Thanks for your work!

Issue #144, opened 2022-07-11
Packaged in distro repos


ErikReider:
I have no clue how this works in any sort of way so help would be very much appreciated. The main distro would be Arch (due to me primarily using it) and maybe Ubuntu and Fedora (I'll probably need to create a fedora build GitHub Action)

lexa:
For building Fedora there is already a rpkg spec file: https://github.com/ErikReider/SwayNotificationCenter/blob/main/swaync.rpkg.spec and a copr repo. I would really appreciate if you could pull it into official Fedora repository though.

ErikReider:
> I would really appreciate if you could pull it into official Fedora repository though

That would be preferred. Thanks @lexa for maintaining the copr repo :)

lexa:
>That would be preferred. 

Sure, but that is rather difficult, especially for such niche package.

alekseifedotov:
The URL will change if we change the maintainer of the Copr package. I don't think there is such thing as *changing the owner* on Fedora copr.

ErikReider:
If you @lexa would be willing to change the maintainer (or any similar option that also changes the URL), I'd be glad to take over it :)

lexa:
@ErikReider Got really excited about this idea of officially maintained Fedora repo and made a mistake of posting from a wrong account.

I've prepared a PR to add Github Action for updating copr repo automatically: https://github.com/ErikReider/SwayNotificationCenter/pull/148

If you want to take over maintenance over copr, you need to get an account on copr.fedorainfracloud.org, as described in the PR.

lexa:
@ErikReider Thanks for making an official copr repo. Could you please enable Fedora 35 builds? Also it would be nice to have aarch64 build.

ErikReider:
Didn't I do that? I'll check

ErikReider:
Should be there if I'm not mistaken:
![image](https://user-images.githubusercontent.com/35975961/180609629-9ad7c088-7ee5-4040-b868-50462c6ad06e.png)

lexa:
sorry, overlooked that configuration.

FilippoBonazziSUSE:
For SUSE distros, work is ongoing in https://build.opensuse.org/package/show/X11:Wayland/SwayNotificationCenter. Hopefully the package will be in Factory (and then Tumbleweed) soon.

werdahias:
[...]

PR #148, opened 2022-07-20
CI: Add Fedora builds (#144)

lexa:
Add Github actions for building SwayNotificationCenter on every push and publish it on Copr when it is released.

`fedora-build.yml` builds the package locally on every commit and PR. It is mostly to check that the build is not broken and there is not missing dependencies.

`fedora-copr.yml` is triggered when SwayNotificationCenter is released. It will start build on the [copr](https://copr.fedorainfracloud.org/). Copr will build the project for different Fedora versions and architectures (x86, arm64, etc).

To make `fedora-copr.yml` work you need to : 
1. Get an account on the [copr](https://copr.fedorainfracloud.org/)
2. Create a new project there, lets call it `SwayNotificationCenter`
3. Set repo secret COPR_REPO_NAME to `SwayNotificationCenter` (same as project name on Copr)
4. Get a token from https://copr.fedorainfracloud.org/api/ and set it in the repo's secrets as 'COPR_API_TOKEN'.

After that, copr shoud be updated when the next release is done.

I see that you don't want to use Github releases, feel free to change trigger event of `fedora-copr.yml` to 
```

on:  
  push:
    tags:
      - '*'
```

### DIFF
--- a/.github/workflows/fedora-build.yml
+++ b/.github/workflows/fedora-build.yml
@@ -1,0 +1,52 @@
+name: Check build for Fedora.
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  package:
+    container: fedora:latest
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install tooling for source RPM build
+        run: |
+          dnf -y install @development-tools @rpm-development-tools
+          dnf -y install rpkg git
+          dnf -y install 'dnf-command(builddep)'
+
+      # It is necessary to checkout into sub-directory, because of some weird ownership problems cause by using containers
+      - name: Check out sources
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          path: swaync
+
+      - name: Generate spec
+        run: |
+          cd swaync
+          mkdir specs
+          rpkg spec --source  --outdir specs
+
+      - name: Install build dependencies
+        run: |
+          cd swaync
+          dnf -y builddep ./specs/swaync.rpkg.spec
+
+      - name: Local build
+        run: |
+          cd swaync
+          mkdir -p out
+          rpkg local --out `pwd`/out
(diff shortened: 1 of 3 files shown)


ErikReider:
Cool! Thanks :)

ErikReider:
Seems to work. I wonder if we can decrease the run time by using a image that already has `@development-tools @rpm-development-tools` installed? (If that even exists)

.github/workflows/fedora-copr.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  package:
+    container: fedora:latest
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install API token for copr-cli
+        env:
+          API_TOKEN_CONTENT: ${{ secrets.COPR_API_TOKEN }}
```
ErikReider, line 18: Add a comment above `# To generate a new token: https://copr.fedorainfracloud.org/api/`. I'm kinda forgetful...

ErikReider (changes requested):

.github/workflows/fedora-build.yml
```diff
@@ -0,0 +1,52 @@
...
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  package:
+    container: fedora:latest
```
ErikReider, line 14: I wonder if we can decrease the run time by using a image that already has @development-tools @rpm-development-tools installed? (If that even exists)

lexa: I've checked the [official fedora docker.hub](https://hub.docker.com/_/fedora) images and there is no docker images with @development-tools installed. Also can't find anything similar at docker.hub.

.github/workflows/fedora-copr.yml
lexa: Done

ErikReider (approved):
LGTM!

ErikReider:
Oh. Forgot about updating the README. I'll fix it :)

PR #161, opened 2022-09-07
Adding instructions for installation on Fedora Silverblue and other rpm-ostree Fedora variants.

foopsss:
Inmutable Fedora systems cannot add COPR repos like the normal Workstation edition as they don't use DNF outside of containers. This pull request addresses this and adds the required instructions for those who may not know.

I'm not exactly sure if repo files for older Fedora versions are taken out once they go EOL, so I just chose to go with the Rawhide repo file since it's exactly the same as the other ones and probably will stay there forever. During the installation, the package corresponding to your Fedora version will be chosen so there should be no problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ dnf install SwayNotificationCenter
 Fedora Silverblue (and other rpm-ostree variants):
 The package can be downloaded from COPR after adding the COPR repo as a ostree repo, and installed as a overlayed package:
 ```zsh
-sudo ostree remote add SwayNotificationCenter https://download.copr.fedorainfracloud.org/results/erikreider/SwayNotificationCenter/fedora-$releasever-$basearch/
+ostree remote add SwayNotificationCenter https://download.copr.fedorainfracloud.org/results/erikreider/SwayNotificationCenter/fedora-$releasever-$basearch/
 rpm-ostree install SwayNotificationCenter
 ```
 
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@ dnf install SwayNotificationCenter
 ```
 
 Fedora Silverblue (and other rpm-ostree variants):
-The package can be downloaded from COPR after manually adding the repo file, and installed as a overlayed package:
+The package can be downloaded from COPR after adding the COPR repo as a ostree repo, and installed as a overlayed package:
 ```zsh
-wget -O /etc/yum.repos.d/erikreider-SwayNotificationCenter.repo https://copr.fedorainfracloud.org/coprs/erikreider/SwayNotificationCenter/repo/fedora-rawhide/erikreider-SwayNotificationCenter-fedora-rawhide.repo
+sudo ostree remote add SwayNotificationCenter https://download.copr.fedorainfracloud.org/results/erikreider/SwayNotificationCenter/fedora-$releasever-$basearch/
 rpm-ostree install SwayNotificationCenter
 ```
 
--- a/README.md
+++ b/README.md
@@ -57,6 +57,13 @@ dnf copr enable erikreider/SwayNotificationCenter
 dnf install SwayNotificationCenter
 ```
 
+Fedora Silverblue (and other rpm-ostree variants):
+The package can be downloaded from COPR after manually adding the repo file, and installed as a overlayed package:
+```zsh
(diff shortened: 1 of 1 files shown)


ErikReider:
Shouldn't there be a better way of doing this? Seems a bit hacky but I've never used Silverblue

foopsss:
> Shouldn't there be a better way of doing this? Seems a bit hacky but I've never used Silverblue

Yeah, it's a bit hacky but that's how it is, since the `yum.repos.d` folder can be written by the user. The only other way to do this works pretty much the same, but uses the `sudo ostree remote add <nameofrepo> <repo file URL>` command, now that I see, which might look fancier for some people.

I've seen both the ´wget´ approach and the ´ostree remote add´ one be recommended to Silverblue users without much difference, and there's an article on Fedora Magazine that recommends both, but I can change the pull request if it makes it better.

Sources:

- https://discussion.fedoraproject.org/t/how-to-add-an-additional-repo-file-on-silverblue/35641
- https://discussion.fedoraproject.org/t/add-external-repos/18951/2
- https://www.reddit.com/r/Fedora/comments/tf5nwu/silverblue_and_installing_driver_from_copr/
- https://fedoramagazine.org/how-i-customize-fedora-silverblue-and-fedora-kinoite/

ErikReider:
I'll look at this tomorrow :)

ErikReider:
The `ostree remote add` version looks a lot better and less "intimidating" so I'd go with that one

foopsss:
Done. The instructions now use `ostree remote add`.

README.md
```diff
@@ -57,6 +57,13 @@ dnf copr enable erikreider/SwayNotificationCenter
 dnf install SwayNotificationCenter
 ```
 
+Fedora Silverblue (and other rpm-ostree variants):
+The package can be downloaded from COPR after adding the COPR repo as a ostree repo, and installed as a overlayed package:
+```zsh
+ostree remote add SwayNotificationCenter https://download.copr.fedorainfracloud.org/results/erikreider/SwayNotificationCenter/fedora-$releasever-$basearch/
```
ErikReider, line 63: Are `$releasever` and `$basearch` already defined in the default fedora env or is this something that the user should fill in?

Also, if you're on version 36, how does updating to a newer version work (eg. 36 -> 37)? Will the repo URL be updated automatically?

foopsss: 1. AFAIK, those variables point the package manager automatically to the version corresponding to your Fedora install. The user does not have to do anything.
2. The repo URL should always work to my understanding because of those variables. **However**, you won't be allowed to upgrade to a new version (like, 36 to 37) if any of your repos doesn't have a package for that new version. The upgrade will stop on its tracks and tell you this, and you'll be required to either remove the outdated repo or wait until that repo gets updated packages. You may also have to uninstall versioned (meant for a specific version) packages, or use packages that aren't versioned (not locked to a specific version). This last bit I just read it from [here](https://discussion.fedoraproject.org/t/simplifying-updates-for-rpm-fusion-packages-and-other-packages-shipping-their-own-rpm-repos) and haven´t actually tried, but I do know from experience that you won't be allowed to upgrade by default if matching packages aren't found.

ErikReider (approved):
LTGM! Thanks :)

cryptoluks:
The command `ostree remote add` is intended specifically for adding OSTree remotes, which is a different concept altogether.

The aforementioned command won't work for the use case of adding DNF repositories, as you've described.

Instead, I recommend using the following command:

```
sudo curl -sL -o /etc/yum.repos.d/erikreider-sway.repo https://copr.fedorainfracloud.org/coprs/erikreider/SwayNotificationCenter/repo/fedora-$(rpm -E %fedora)/erikreider-SwayNotificationCenter-fedora-$(rpm -E %fedora).repo && \
rpm-ostree install SwayNotificationCenter
```

Regarding the information from the Fedora discussion posts, it appears to be inaccurate. The author himself admitted that he never actually attempted the process, suggesting that his answer is primarily based on speculation rather than first-hand experience or concrete knowledge.

ErikReider:
Looking into it further you seem to be correct. I'll update the README, thanks! :)

ErikReider:
Fixed in 9824048

PR #170, opened 2022-10-07
feat: transient notification visibility

lazytanuki:
Hi and thanks for this great tool !

This PR adds a new **configurable** visibility state called transient. It allows the user to set the transient state of a notification in order for it to show, but not to be added to the CC.
This is useful, for instance, for certain music apps that show a notification on track changes. We might want to see them, but we don't necessarily want them in the CC.

Thanks !

### DIFF
--- a/man/swaync.5.scd
+++ b/man/swaync.5.scd
@@ -110,7 +110,7 @@ config file to be able to detect config errors
 			type: string ++
 			optional: false ++
 			default: enabled ++
-			values: ignored, muted, enabled ++
+			values: ignored, muted, transient, enabled ++
 			description: The notification visibility state. ++
 		*app-name*++
 			type: string ++
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -137,7 +137,8 @@ namespace SwayNotificationCenter {
     public enum NotificationStatusEnum {
         ENABLED,
         MUTED,
-        IGNORED;
+        IGNORED,
+        TRANSIENT;
 
         public string to_string () {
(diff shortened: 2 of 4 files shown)


ErikReider (approved):
LGTM! Thanks for the help :)

StayPirate:
I set the `transient` state on one rule and I noticed that subsequent scripts are not ran. I that an intended behavior, a bug, or did I make an error on my configuration?

```json
  "notification-visibility": {
    "slack-show-only": {
      "_comment": "Show Slack messages but do not add them to the CC",
      "state": "transient",
      "urgency": "Low",
      "app-name": "Firefox",
      "summary": "New message (in|from) .*"
    }
  },
  "scripts": {
    "_comment": "From swaync(5): Only the first matching script will be run.",
    "_comment2": "In order to sort rules letters need to be used, digits don't work... So, AA is loaded before AB.",
    "AB-slack-sound": {
      "_comment": "Play a notification sounds for Slack notification from FF",
      "exec": "slack-notification-alert",
      "app-name": "Firefox",
      "summary": ".*New message (in #|from ).*"
    }
  }
```

ErikReider:
@StayPirate What's the output of swaync if you run `G_MESSAGES_DEBUG=all swaync` in the terminal?

StayPirate:
Never mind, the problem I faced was triggered from [a bug](https://github.com/StayPirate/dotfiles/commit/0630295cac33ed996768d0bd961a7e541794aad1) on one of my bash scripts. Hence, please ignore my previous comment.

PR #175, opened 2022-10-18
Add option to configure layer for control center window

junglerobba:
Hi!

This adds a config option to configure the layer_shell layer, ie the rules for above/below which windows the control center will appear. This is mostly for overlay so it can appear above fullscreen windows, I doubt background and bottom are useful really, but they're there, so why not expose them :shrug:

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -42,6 +42,25 @@ namespace SwayNotificationCenter {
         }
     }
 
+    public enum Layer {
+        BACKGROUND, BOTTOM, TOP, OVERLAY;
+
+        public string parse () {
+            switch (this) {
+                default:
+                    return "top";
+                case BACKGROUND:
+                    return "background";
+                case BOTTOM:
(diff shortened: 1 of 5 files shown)


src/configModel/configModel.vala
```diff
@@ -42,6 +42,23 @@ namespace SwayNotificationCenter {
...
+                default:
+                    return "top";
+                case BACKGROUND:
+                    return "background";
+                case TOP:
+                    return "top";
+                case OVERLAY:
+                    return "overlay";
+            }
+        }
```
ErikReider, line 59: `BOTTOM` is missing

ErikReider (changes requested):

src/configModel/configModel.vala
junglerobba: Yep, somehow missed this

ErikReider (approved):
LGTM! I really appreciate the help! :)

PR #190, opened 2022-12-23
Update README with correct config man page

nichobi:
The config details are in `swaync(5)` not `(1)` as currently stated. `swaync(1)` only contains the same information (though in `(1)` it correctly references `(5)` for details)


ErikReider (approved):
Thanks a ton!

PR #194, opened 2023-01-01
Action scripts

### LINKED ISSUE #192: [Feature] Custom action on click
**Please read through the README and the Man pages before submitting**
**Please also make sure that there isn't any prior issue describing this feature**

**Is your feature request related to a problem? Please describe.**
Discord notifications have a feature where if you click on them, it brings you to the message / the channel; the action is set to something to allow you to view it. Unfortunately, it doesn't focus the workspace that I have Discord open in.

**Describe the solution you'd like**
If I could just set a custom action to run (with the default action provided by Discord still working), I could make my workspace switch to the one I have Discord open in.
This would ideally look similar to how the scripts on receiving notifications work, e.g.
```json
{
  "actions": {
    // This script will run when clicking on a Discord notification, bringing you
    // to the third Hyprland workspace (where I put Discord)
    "example-script": {
      "exec": "hyprctl dispatch workspace 3",
      "app-name": "discord"
    }
  }
}
```

**Describe alternatives you've considered**
I can't think of any other alternative. I could just not have the notification switch my workspace to Discord, but that's kind of annoying and not really good UX.

auctumnus:
Closes #192.

This solution is likely not the most maintainable. `run_action_scripts` duplicates most of the content from `new_notification`, and requires you to repeat a lot of the `NotifyParams` content. Unfortunately, I'm not familiar enough with Vala to know how to fix this. (That is - this is my first time programming in Vala, so please check over all of my work.)

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -204,12 +204,12 @@ namespace SwayNotificationCenter {
         public NotificationStatusEnum state { get; set; }
     }
 
-#if WANT_SCRIPTING
     public enum ScriptRunOnType {
         ACTION,
         RECEIVE;
     }
 
+#if WANT_SCRIPTING
     public class Script : NotificationMatching {
         public string ? exec { get; set; default = null; }
         public ScriptRunOnType run_on { get; set; default = ScriptRunOnType.RECEIVE; }
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -212,9 +212,7 @@ namespace SwayNotificationCenter {
 
     public class Script : NotificationMatching {
         public string ? exec { get; set; default = null; }
-        public string ? run_on { get; set; default = "receive"; }
-        // TODO: this should be an enum instead!
-        //public ScriptRunOnType ? run_on { get; set; default = ScriptRunOnType.RECEIVE; }
+        public ScriptRunOnType run_on { get; set; default = ScriptRunOnType.RECEIVE; }
 
         public async bool run_script (NotifyParams param, out string msg) {
             msg = "";
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -198,10 +198,6 @@ namespace SwayNotificationCenter {
                     return TRANSIENT;
             }
         }
-
-        public string parse () {
-            return this.to_string();
-        }
     }
 
     public class NotificationVisibility : NotificationMatching {
@@ -209,9 +205,16 @@ namespace SwayNotificationCenter {
     }
 
 #if WANT_SCRIPTING
+    public enum ScriptRunOnType {
+        ACTION,
+        RECEIVE;
+    }
+
     public class Script : NotificationMatching {
         public string ? exec { get; set; default = null; }
         public string ? run_on { get; set; default = "receive"; }
+        // TODO: this should be an enum instead!
+        //public ScriptRunOnType ? run_on { get; set; default = ScriptRunOnType.RECEIVE; }
 
         public async bool run_script (NotifyParams param, out string msg) {
             msg = "";
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -198,6 +198,10 @@ namespace SwayNotificationCenter {
                     return TRANSIENT;
             }
         }
+
+        public string parse () {
+            return this.to_string();
+        }
     }
 
     public class NotificationVisibility : NotificationMatching {
@@ -207,6 +211,7 @@ namespace SwayNotificationCenter {
 #if WANT_SCRIPTING
     public class Script : NotificationMatching {
         public string ? exec { get; set; default = null; }
+        public string ? run_on { get; set; default = "receive"; }
 
         public async bool run_script (NotifyParams param, out string msg) {
             msg = "";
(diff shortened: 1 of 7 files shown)


src/notiDaemon/notiDaemon.vala
```diff
@@ -242,6 +243,85 @@ namespace SwayNotificationCenter {
...
+                                           e.message);
+                        }
+                    }
+                });
+                break;
+            }
+#endif
+        }
+
+
```
ErikReider, line 324: You could always just combine the two into a new method :)

src/notiDaemon/notiDaemon.vala
```diff
@@ -242,6 +243,85 @@ namespace SwayNotificationCenter {
...
+        public void run_action_scripts(
+            uint32 id,
+            string app_name,
+            uint32 replaces_id,
+            string app_icon,
+            string summary,
+            string body,
+            string[] actions,
+            HashTable<string, Variant> hints,
+            int expire_timeout) throws DBusError, IOError {
```
ErikReider, line 257: To make your life easier, I'd recommend only requiring a `NotifyParams` when calling the method :)

src/notification/notification.vala
```diff
@@ -287,13 +287,44 @@ namespace SwayNotificationCenter {
...
+                    param.app_icon,
+                    param.summary,
+                    param.body,
+                    new string[0],
+                    param.hints,
+                    param.expire_timeout
+                );
+            } catch (Error e) {
+                print ("Error: %s\n", e.message);
+            }
```
ErikReider, line 327: Use the `action_clicked` method instead. Less duplicated code

src/configModel/configModel.vala
```diff
@@ -207,6 +211,7 @@ namespace SwayNotificationCenter {
 #if WANT_SCRIPTING
     public class Script : NotificationMatching {
         public string ? exec { get; set; default = null; }
+        public string ? run_on { get; set; default = "receive"; }
```
ErikReider, line 214: I'd prefer the use of an enum. It'll make it easier to keep track of in the future

ErikReider (changes requested):
Otherwise it looks good to me! :)

src/notiDaemon/notiDaemon.vala
ErikReider: Something like this:

```vala
[DBus (visible = false)]
public void run_action_scripts(NotifyParams param) throws DBusError, IOError {
```

The `[DBus (visible = false)]` ensures that the method won't be exported to the DBUS interface so using `NotifyParams` works as expected

auctumnus: Oh, I didn't know about that! That should solve most of the egregious issues - I'll go ahead and fix that up.

src/configModel/configModel.vala
```diff
@@ -198,6 +198,10 @@ namespace SwayNotificationCenter {
                     return TRANSIENT;
             }
         }
+
+        public string parse () {
+            return this.to_string();
+        }
```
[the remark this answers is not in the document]
auctumnus: Oops, I'm not sure why that got in there. I think it was from attempting to use an enum for the run_on type originally, and getting confused as to why there were compile errors? I'll remove that.

auctumnus:
I tried getting it to be an enum, but I can't figure out how to get it to serialize the enum value from the config? Otherwise, the other issues are fixed. It _should_ be a pretty simple change to have it not be a string - I have the enum there in the config model and such.

src/notiDaemon/notiDaemon.vala
```diff
@@ -238,10 +257,17 @@ namespace SwayNotificationCenter {
...
+        }
 
-            return id;
+        /**
+         * Runs scripts that are set to run on action.
+         */
+        [DBus (visible = false)]
+        public void run_action_scripts (NotifyParams param) throws DBusError, IOError {
+            this.run_scripts (param, "action");
         }
```
ErikReider, line 268: I'd like this merged into one method. Makes it easier to read and understand :)

ErikReider (changes requested):

ErikReider:
> I tried getting it to be an enum, but I can't figure out how to get it to serialize the enum value from the config? Otherwise, the other issues are fixed. It _should_ be a pretty simple change to have it not be a string - I have the enum there in the config model and such.

Removing the question mark from `public ScriptRunOnType ? run_on { get; set; default = ScriptRunOnType.RECEIVE; }` should do the trick

auctumnus:
Both of those should now be fixed. Thanks for the tip about removing the question mark!

src/notiDaemon/notiDaemon.vala
```diff
@@ -195,9 +195,28 @@ namespace SwayNotificationCenter {
             // Run the first script if notification meets requirements
             HashTable<string, Script> scripts = ConfigModel.instance.scripts;
             if (scripts.length == 0) return id;
+            this.run_scripts (param, ScriptRunOnType.RECEIVE);
```
ErikReider, line 198: Missing `#endif`

src/configModel/configModel.vala
```diff
@@ -205,8 +205,14 @@ namespace SwayNotificationCenter {
     }
 
 #if WANT_SCRIPTING
+    public enum ScriptRunOnType {
+        ACTION,
+        RECEIVE;
+    }
+
```
ErikReider, line 212: Should be outside of `#if WANT_SCRIPTING`

ErikReider (changes requested):
Sorry, some last minute changes are needed

auctumnus:
Should be fixed; thanks for spotting these, I hadn't tried building without scripting so I wouldn't have seen these.

ErikReider (approved):
Thanks for the help! ❤️

auctumnus:
No problem! Thank you for helping me through this PR :+1:

PR #199, opened 2023-01-13
Add Override urgency to config

mildred:
Allow to override the urgency of some notifications from the config file

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -200,8 +200,35 @@ namespace SwayNotificationCenter {
         }
     }
 
+    public enum NotificationUrgencyEnum {
+        UNSET = -1,
+        LOW = UrgencyLevels.LOW,
+        NORMAL = UrgencyLevels.NORMAL,
+        CRITICAL = UrgencyLevels.CRITICAL;
+
+        public uint8 to_byte () {
+            return this;
+        }
+
+        public string to_string () {
+            switch (this) {
+                case LOW:
+                    return "Low";
+                case NORMAL:
+                    return "Normal";
+                case CRITICAL:
+                    return "Critical";
+                default:
+                    return "Unset";
+            }
+        }
+    }
+
     public class NotificationVisibility : NotificationMatching {
         public NotificationStatusEnum state { get; set; }
+        public NotificationUrgencyEnum override_urgency {
+            get; set; default = NotificationUrgencyEnum.UNSET;
+        }
     }
 
     public enum ScriptRunOnType {
--- a/src/configSchema.json
+++ b/src/configSchema.json
@@ -224,6 +224,12 @@
               "default": "Normal",
               "enum": ["Low", "Normal", "Critical"]
             },
+            "override-urgency": {
+              "type": "string",
+              "description": "The new urgency of the notification (optional)",
+              "default": "unset",
+              "enum": ["unset", "low", "normal", "critical"]
+            },
             "category": {
               "type": "string",
               "description": "Which category the notification belongs to. Uses Regex."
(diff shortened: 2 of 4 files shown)


ErikReider:
Thanks for the PR! Request a review from me when you're done with the draft :)

src/configModel/configModel.vala
```diff
@@ -200,8 +200,47 @@ namespace SwayNotificationCenter {
         }
     }
 
+    public enum NotificationUrgencyEnum {
+        UNSET = -1,
+        LOW = 0,
+        NORMAL = 1,
+        CRITICAL = 2;
+
+        public static NotificationUrgencyEnum from_value (string value) {
```
mildred, line 209: For some reason, I believe that this function is never called by the JSON deserialization process (I tried to run `value.ascii_down()` but it was not called). Same for the enum just above.

ErikReider: The JSON parser checks this automatically. I created such methods for other enums for use in other methods

src/configSchema.json
```diff
@@ -224,6 +224,12 @@
               "default": "Normal",
               "enum": ["Low", "Normal", "Critical"]
             },
+            "override-urgency": {
+              "type": "string",
+              "description": "The new urgency of the notification (optional)",
+              "default": "Unset",
```
ErikReider, line 230: `Unset` should be lower-case just like below

src/configSchema.json
```diff
@@ -224,6 +224,12 @@
...
+              "enum": ["unset", "low", "normal", "critical"]
+            },
```
ErikReider, line 232: Make sure to update the Man(5) page

src/configModel/configModel.vala
```diff
@@ -200,8 +200,47 @@ namespace SwayNotificationCenter {
         }
     }
 
+    public enum NotificationUrgencyEnum {
+        UNSET = -1,
+        LOW = 0,
+        NORMAL = 1,
+        CRITICAL = 2;
```
ErikReider, line 207: Use the `UrgencyLevels` enums as values for these. Ex: `LOW = UrgencyLevels.LOW,`

src/configModel/configModel.vala
```diff
@@ -200,8 +200,47 @@ namespace SwayNotificationCenter {
...
+                case "low":
+                    return LOW;
+                case "normal":
+                    return NORMAL;
+                case "critical":
+                    return CRITICAL;
+                default:
+                    return UNSET;
+            }
+        }
```
ErikReider, line 220: Unused, should be removed

src/configModel/configModel.vala
```diff
@@ -200,8 +200,47 @@ namespace SwayNotificationCenter {
...
+                case UNSET:
+                    return "Unset";
+            }
+            return "Unset";
+        }
+    }
+
     public class NotificationVisibility : NotificationMatching {
         public NotificationStatusEnum state { get; set; }
+        public NotificationUrgencyEnum override_urgency { get; set; default = NotificationUrgencyEnum.UNSET; }
```
ErikReider, line 243: I'd prefer if this wasn't on a single line. Like this:
```vala
public NotificationUrgencyEnum override_urgency {
    get; set; default = NotificationUrgencyEnum.UNSET;
}
```

src/configModel/configModel.vala
```diff
@@ -200,8 +200,47 @@ namespace SwayNotificationCenter {
...
+                case LOW:
+                    return "Low";
+                case NORMAL:
+                    return "Normal";
+                case CRITICAL:
+                    return "Critical";
+                case UNSET:
+                    return "Unset";
+            }
+            return "Unset";
```
ErikReider, line 237: Join these with a `default`. Fewer duplicated strings :)

ErikReider (changes requested):
Other than that, LGTM! :)

ErikReider (approved):
LGTM. Thanks for the help! :)

PR #200, opened 2023-01-13
Add COPYING and README to pkgbuild install

bd-g:
Per #144 to prep for trying to get the package in \[community\]

COPYING is standard naming for license files under GNU, so no need to change it (counter to my comment on #144)


ErikReider (approved):
LGTM! Thanks for the quick fix! :)

PR #201, opened 2023-01-18
Additional Widgets

JannisPetschenka:
Adding extra widgets to the controll center like brightness & audio controll and power buttons


JannisPetschenka:
I'm new to vala and this project so I have a couple of issues where I'm don't the best solution :)

Main one being: the values for the brightness and audio control get set at the beginning, when starting the daemon. If you change the values via the cli or kb-buttons it won't be updated in the control center. I thought about adding a timer to the Widgets to check every X seconds, but I don't know if that is the best solution. 

Also I thought about changing the brightness and audio widget to allow other cli tools than brightnessctl and pamixer...

ErikReider:
> Main one being: the values for the brightness and audio control get set at the beginning, when starting the daemon. If you change the values via the cli or kb-buttons it won't be updated in the control center. I thought about adding a timer to the Widgets to check every X seconds, but I don't know if that is the best solution.

Probably the best solution would be to tie the calls and listen to DBus/pulsedaemon. I've got a pulsedaemon example here: [SwaySettings Pulse page](https://github.com/ErikReider/SwaySettings/tree/main/src/Pages/Pulse). The brightness should be accessible from DBus if I remember correctly.

The reason for going directly for the source of those two is that it's easier to change and listen to changes :)

> Also I thought about changing the brightness and audio widget to allow other cli tools than brightnessctl and pamixer...

Wouldn't be needed when directly accessing the daemons

If you have any other questions or need help, just ask :)

ErikReider:
How easy would it be to split each widget into their own PR? Would make both of our lives a lot easier and focused 😅

README.md
```diff
@@ -49,6 +49,72 @@ These widgets can be customized, added, removed and even reordered
...
+    "controll-buttons": [
+      {
+        "label": "直",
+        "command": "~/.config/rofi/rofi-wifi-menu.sh"
+      },
+      {
+        "label": "",
+        "command": "~/.config/rofi/rofi-bluetooth"
+      }
+    ]
```
ErikReider, line 113: Similar to what I wrote above but with the `type` set to `button`?

JannisPetschenka:
> Probably the best solution would be to tie the calls and listen to DBus/pulsedaemon. I've got a pulsedaemon example here: [SwaySettings Pulse page](https://github.com/ErikReider/SwaySettings/tree/main/src/Pages/Pulse). The brightness should be accessible from DBus if I remember correctly.

Aaah, I knew there had to be a better solution :) Thanks, I'll look into it 👍

PR #206, opened 2023-01-30
Bump minimal meson version to 0.59.0, replace 'get_pkgconfig_variable'

freijon:
The minimal meson version is too low, features from newer versions are used:

```
WARNING: Project specifies a minimum meson_version '>= 0.50.0' but uses features which were added in newer versions:
 * 0.54.0: {'message with more than one argument'}
 * 0.59.0: {'feed arg in custom_target'}
NOTICE: Future-deprecated features used:
 * 0.56.0: {'dependency.get_pkgconfig_variable'}
```


ErikReider (approved):
LGTM! Thanks! :)

PR #207, opened 2023-02-09
Audio slider

JannisPetschenka:
Got the next PR already prepared :) It should be a bit more straight forward since it's pretty much just your code from [SwaySettings](https://github.com/ErikReider/SwaySettings/tree/main/src/Pages/Pulse) with some stuff removed ;)


man/swaync.5.scd
```diff
@@ -366,6 +366,16 @@ config file to be able to detect config errors
...
+		*volume*++
+			type: object ++
+			css class: widget-volume ++
+			properties: ++
+				label: ++
+					type: string ++
+					optional: true ++
+					default: "Volume" ++
+					description: Text displayed in front of the volume slider ++
+			description: Slider to control pulse volume ++
```
ErikReider, line 378: Remove extra tabs on line 378 to fix build error

JannisPetschenka: Thx. Already found the error with `libpulse` not being found and the lint issue :) Just looking if I find some more ;)

ErikReider: I'll have to rebuild the docker images on my end for the build actions to work. I wonder if GitHub can do this for me? 🤔

JannisPetschenka: Not familiar with GitHub on this (only used GitLab CI/CD until now). But I think you should be able to run `docker build` you might have to add docker as a dependency for the action though.

ErikReider:
Hmm... Maybe I'll just move the installation of the needed packages into the respective action. Seems easier than hosting a docker image that I usually forget to update

JannisPetschenka:
> Hmm... Maybe I'll just move the installation of the needed packages into the respective action. Seems easier than hosting a docker image that I usually forget to update

True, might be easier.. But here is a way to do it via an action [GitHub Docker action](https://thedatabaseme.de/2022/07/02/building-and-pushing-use-github-actions-to-build-and-push-docker-images/).. I could look into it a bit more next weekend if you'd like.

ErikReider:
> True, might be easier.. But here is a way to do it via an action [GitHub Docker action](https://thedatabaseme.de/2022/07/02/building-and-pushing-use-github-actions-to-build-and-push-docker-images/).. I could look into it a bit more next weekend if you'd like.

Wouldn't the main image be updated then? Or can you create different sub-tags and delete them on merge?

JannisPetschenka:
> Wouldn't the main image be updated then? Or can you create different sub-tags and delete them on merge?

Hmm true.. you could build/use the images locally in the action when on sub-branches and only push to docker-hub when on main. But I think that would still be a bad idea since that would mean you'd push the images everytime you push to main. Even if there were no changes to the dockerfiles (which I assume is the case 90% of the time).
So manually might be the way to go :)

src/controlCenter/widgets/volume/pulseDevice.vala
```diff
@@ -0,0 +1,134 @@
+using PulseAudio;
```
ErikReider, line 1: We should add a perma link to the SwaySettings file as a comment at the top. Something like `// From SwaySettings PulseAudio page: PERMALINK`

src/controlCenter/widgets/volume/pulseDaemon.vala
```diff
@@ -0,0 +1,530 @@
+using PulseAudio;
```
ErikReider, line 1: Permalink here too

src/controlCenter/widgets/volume/volume.vala
```diff
@@ -0,0 +1,65 @@
+using GLib;
```
ErikReider, line 1: GLib is used by default, so shouldn't be needed :)

README.md
```diff
@@ -51,6 +51,7 @@ These widgets can be customized, added, removed and even reordered
 - Mpris (Media player controls for Spotify, Firefox, Chrome, etc...)
 - Menubar with dropdown and buttons
 - Button grid
+- Volume slider using pulse audio
```
ErikReider, line 54: Should be "PulseAudio"

src/controlCenter/widgets/volume/volume.vala
```diff
@@ -0,0 +1,65 @@
...
+
+        string label = "Volume";
+
+        private PulseDevice ? default_sink = null;
+        private PulseDaemon client = new PulseDaemon ();
+
+        construct {
+            this.client.change_default_device.connect (default_device_changed);
+
+            slider.value_changed.connect (() => {
```
ErikReider, line 22: Check if `default_sink` is null

src/controlCenter/widgets/volume/volume.vala
```diff
@@ -0,0 +1,65 @@
...
+            });
+        }
+
+
+        public Volume (string suffix, SwayncDaemon swaync_daemon, NotiDaemon noti_daemon) {
+            base (suffix, swaync_daemon, noti_daemon);
+
+            Json.Object ? config = get_config (this);
+            if (config != null) {
+                string ? l = get_prop<string> (config, "label");
```
ErikReider, line 36: Set the label text directly and remove the `label` variable

src/controlCenter/widgets/volume/volume.vala
```diff
@@ -0,0 +1,65 @@
...
+
+namespace SwayNotificationCenter.Widgets {
+    public class Volume : BaseWidget {
+        public override string widget_name {
+            get {
+                return "volume";
+            }
+        }
+
+        Gtk.Label label_widget;
```
ErikReider, line 11: Construct with `null`

ErikReider (changes requested):
I'll have to remove the dockerfiles in main to fix the build issues. Other than that, the requested changes, and if everything builds correctly, I'll merge :)

man/swaync.5.scd
ErikReider: Merge from main and it should be fixed :)

JannisPetschenka:
Fixed the issues and added the additional dependencies to the build files, should hopefully work now :)

src/controlCenter/widgets/volume/volume.vala
```diff
@@ -0,0 +1,62 @@
...
+
+
+        public Volume (string suffix, SwayncDaemon swaync_daemon, NotiDaemon noti_daemon) {
+            base (suffix, swaync_daemon, noti_daemon);
+
+            Json.Object ? config = get_config (this);
+            if (config != null) {
+                string ? l = get_prop<string> (config, "label");
+                if (l != null) label_widget.set_label (l);
+                else label_widget.set_label ("Volume");
```
ErikReider, line 36: You should be able to simplify this to:

```vala
string ? label = get_prop<string> (config, "label");
label_widget.set_label (label ?? "Volume");
```

JannisPetschenka: Done :)

ErikReider (approved):
LGTM :)

PR #211, opened 2023-02-12
Add Brightness slider widget

JannisPetschenka:
And here the third PR :) not quite ready yet, have to implement setting the brightness, but that should work via the `org.freedesktop.login1` dbus (haven't tested it yet though) :grin: 

Didn't find a way to get the brightness via DBus without using something Gnome, KDE or else specific, so I'm just using a file monitor :)


JannisPetschenka:
Added the set_brightness functionality :) 
There are two things I still want to look at before it's ready. 

1. add possibility to set a min-cap to the brightness because accidentally setting the brightness to 0 could be annoying
2. maybe allow changing the brightness of devices in the `/sys/class/leds` folder (keyboard-backlight etc). Not quite sure how useful that would be though.. and I might have to change the slider for that since the value for keyboards is usually between 0 and 3 and that could get messy with the conversion :)

JannisPetschenka:
Done :) The slider for the kb-brightness is a bit janky since you can seamlessly set values between 0 and 100 but the brightness will only change on certain thresholds because the kb-backlight doesn't have that many brightness levels :) I thought about changing the range of the scale but that would be a bit annoying (but possible) since not all keyboard have the same amount of brightness-steps. :upside_down_face:

JannisPetschenka:
> Done :) The slider for the kb-brightness is a bit janky since you can seamlessly set values between 0 and 100 but the brightness will only change on certain thresholds because the kb-backlight doesn't have that many brightness levels :) I thought about changing the range of the scale but that would be a bit annoying (but possible) since not all keyboard have the same amount of brightness-steps. upside_down_face
> 
> Edit: I will look into the issue with the kb-brightness slider again next weekend. Might be better to set the Max-value depending on the max level of the KB :)

Fixed :) Should be ready now :D 
Also added a hint on how to address specific widgets with a name after `#` since I couldn't find it in the docs and had to look at the code to figure it out :)

src/configSchema.json
```diff
@@ -430,6 +430,34 @@
           "default": "Volume"
         }
       }      
+    },
+    "backlight": {
```
ErikReider, line 434: Add to `widget-config` object

src/controlCenter/widgets/backlight/backlightUtil.vala
```diff
@@ -0,0 +1,113 @@
...
+                int val = calc_percent (int.parse (data));
+                this.brightness_change (val);
+            } catch (Error e) {
+                error ("Error %s\n", e.message);
+            }
+        }
+
+        private void set_max_value () {
+            try {
+
```
ErikReider, line 91: Remove extra line

src/controlCenter/widgets/backlight/backlight.vala
```diff
@@ -0,0 +1,64 @@
...
+            base (suffix, swaync_daemon, noti_daemon);
+
+            Json.Object ? config = get_config (this);
+            if (config != null) {
+                string ? label = get_prop<string> (config, "label");
+                label_widget.set_label (label ?? "Brightness");
+                string ? device = get_prop<string> (config, "device");
+                string ? subsystem = get_prop<string> (config, "subsystem");
+                if (subsystem != "backlight" && subsystem != "leds") {
+                    info ("Invalid subsystem for device %s. Use 'backlight' or 'leds'. Using default: 'backlight'", device);
```
ErikReider, line 26: Possibility of printing NULL when device isn't defined

src/controlCenter/widgets/backlight/backlight.vala
```diff
@@ -0,0 +1,64 @@
...
+                    subsystem = "backlight";
+                }
+                client = new BacklightUtil (subsystem ?? "backlight", device ?? "intel_backlight");
```
ErikReider, line 29: Move the null check to the `get_prop` lines (`get_prop (...) ?? "intel_backlight"`)

src/controlCenter/widgets/backlight/backlight.vala
```diff
@@ -0,0 +1,64 @@
...
+
+                int ? min = get_prop<int> (config, "min");
+                if (min == null) min = 0;
```
ErikReider, line 32: Remove null check. Null int defaults to 0. Also remove the `?` from `int ?`

src/controlCenter/widgets/backlight/backlight.vala
ErikReider, line 26: Convert this into a switch statement with the checks in line 33-34 included and this check into `default`.

```vala
switch (subsystem) {
    default:
    case "backlight":
        info ("Invalid subsystem for device %s. " +
              "Use 'backlight' or 'leds'. Using default: 'backlight'",
              device);
        subsystem = "backlight";
        slider.set_range (min, 100);
        break;
    case "leds":
        slider.set_range (min, this.client.get_max_value ());
        break;
}
```

src/controlCenter/widgets/backlight/backlight.vala
```diff
@@ -0,0 +1,64 @@
...
+
+            add (label_widget);
+            pack_start (slider, true, true, 0);
+
+            show_all ();
+        }
+
+        public void brightness_changed (int percent) {
+            slider.set_value (percent);
+        }
```
ErikReider, line 54: Move into lambda function `brightness_change.connect ((percent) {...});`

src/controlCenter/widgets/backlight/backlightUtil.vala
```diff
@@ -0,0 +1,113 @@
+namespace SwayNotificationCenter.Widgets {
+    class BacklightUtil {
+
+        [DBus (name = "org.freedesktop.login1.Session")]
+        interface Login1 : Object {
+            public abstract void set_brightness (string subsystem, string name, uint32 brightness) throws GLib.Error;
```
ErikReider, line 6: Line limit

src/controlCenter/widgets/backlight/backlightUtil.vala
```diff
@@ -0,0 +1,113 @@
...
+        string subsystem;
+
+        public signal void brightness_change (int percent);
+
+        public BacklightUtil (string s, string d) {
+            this.subsystem = s;
+            this.device = d;
+
+            path_current = Path.build_filename ("/sys/class/" + subsystem + "/" + device + "/brightness");
+            path_max = Path.build_filename ("/sys/class/" + subsystem + "/" + device + "/max_brightness");
```
ErikReider, line 27: Use `Path.build_path` instead:
```vala
path_current = Path.build_path (Path.DIR_SEPARATOR_S,
                                "sys", "class", subsystem, device, "brightness");
```

src/controlCenter/widgets/backlight/backlightUtil.vala
```diff
@@ -0,0 +1,113 @@
...
+            fd = File.new_for_path (path_current);
```
ErikReider, line 28: Check if file exists with `fd.query_exists ()`. Currently segfaults when the file doesn't exist so don't run `set_max_value` if it doesn't exist. The widget should be hidden/removed if there's no device found

src/controlCenter/widgets/backlight/backlightUtil.vala
```diff
@@ -0,0 +1,113 @@
...
+
+        [DBus (name = "org.freedesktop.login1.Session")]
+        interface Login1 : Object {
+            public abstract void set_brightness (string subsystem, string name, uint32 brightness) throws GLib.Error;
+        }
+
+        string path_current;
+        string path_max;
+        File fd;
+        FileMonitor monitor;
```
ErikReider, line 12: Should be null as default and should be checked if null due to the chance that the try catch statement below might fail

ErikReider (changes requested):

src/controlCenter/widgets/backlight/backlightUtil.vala
JannisPetschenka: What do you mean here? :upside_down_face:

ErikReider:
Oh, just saw the email from GitHub! I'll review this tomorrow! :)

src/controlCenter/widgets/backlight/backlight.vala
```diff
@@ -0,0 +1,73 @@
...
+        public Backlight (string suffix, SwayncDaemon swaync_daemon, NotiDaemon noti_daemon) {
+            base (suffix, swaync_daemon, noti_daemon);
+
+            Json.Object ? config = get_config (this);
+            if (config != null) {
+                string ? label = get_prop<string> (config, "label");
+                label_widget.set_label (label ?? "Brightness");
+                string device = (get_prop<string> (config, "device") ?? "intel_backlight");
+                string subsystem = (get_prop<string> (config, "subsystem") ?? "backlight");
+                int min = get_prop<int> (config, "min");
```
ErikReider, line 25: Adding a int.min() to would ensure that it won't go under 0: `int min = int.max(0, get_prop<int>(config, "min"));`

ErikReider (changes requested):
Just a small change. Other than that, LGTM!

ErikReider (approved):
LGTM! Thanks :)

Issue #179, opened 2022-11-14
Handle action of type text


xfalcox:
On Windows/Android/ChromeOS, when you send a notification that contains an action of type text, it will allow the user to type some text in a input box and send the action with the typed text as the parameter.

Android:
![image](https://user-images.githubusercontent.com/1385470/201687917-a3d142dc-2064-4fdb-8cc5-c1d8077c72ab.png)

Windows:
![image](https://user-images.githubusercontent.com/1385470/201687968-d8e14b03-0c1d-463b-818c-b9f52725b472.png)

When using snc there is a button for the action with type text, but clicking on it send the notification with a `null` value for the action.

This is useful, for example, in chat apps so you can type a quick response without having to open a window of the app.

ErikReider:
is there a application in Linux that sends these kind of notifications?

ErikReider:
> Chrome does.

Which website?

Could you run swaync with all messages enabled (`G_MESSAGES_DEBUG=all swaync`), receive that notification and paste the output here?

xfalcox:
> Which website?

For testing you can use https://tests.peter.sh/notification-generator/ and select "Actions = One action (type text with placeholder)" to trigger it as you want.

If you add a breakpoint to the sw.js file on your Chrome instance DevTools on line 23 and inspect the `event` object, you will see that the `reply` attribute value is null. That's where the text would be inserted.


> Could you run swaync with all messages enabled (G_MESSAGES_DEBUG=all swaync), receive that notification and paste the output here? 

Sure, here it is:

```
** (swaync:94956): DEBUG: 16:31:38.166: notiDaemon.vala:153: Notification: 
category:		(null)
image_data:		false
expire_timeout:		-1
replaces_id:		0
desktop_entry:		chromium
body:			tests.peter.sh

Notification content
app_icon:		file:///tmp/.org.chromium.Chromium.IfUXnn
action_icons:		false
time:			1668454298
hints:		
	urgency: uint32 1
	desktop-entry: 'chromium'
	image_path: data
	image-path: data
applied_id:		1
image_path:		/tmp/.org.chromium.Chromium.y3MoXf
app_name:		Chromium
summary:		Notification title
icon_data:		false
resident:		false
default_action:		Name: Activate, Id: default
actions:		
	Name: Reply, Id: 0
	Name: Settings, Id: settings
urgency:		Low
```

xfalcox:
[...]

PR #221, opened 2023-02-21
Add inline replies to notifications

ErikReider:
Fixes: #179 

Had to do a lot of tweaking in this PR due to a Notification essentially being a GtkButton with content inside of it. I replaced said GtkButton with an GtkEventBox that behaves like a button but respects children with input.

Todo:
- [ ] Double check the CSS and make it more accessible (probably contrast)
- [ ] Make use of `KeyboardMode.ON_DEMAND` to be able to focus the GtkEntry when in NotiWindow. (Can't test in Sway. Need to test somewhere else?)
- [ ] Test in applications

image


tukusejssirs:
@ErikReider, do I get it right that inline replies will work on Sway after swaywm/sway#7587 is merged? 🤔

ErikReider:
> @ErikReider, do I get it right that inline replies will work on Sway after [swaywm/sway#7587](https://github.com/swaywm/sway/pull/7587) is merged? thinking

Will work now, but only in the control center. We check if the compositor supports the ON_DEMAND interactivity which is required to get keyboard focus on floating notifications which is added in that PR :)

PR #230, opened 2023-03-03
Menu animation

JannisPetschenka:
Small changes to the menubar to add some animation :)

### DIFF
--- a/src/configSchema.json
+++ b/src/configSchema.json
@@ -423,6 +423,17 @@
               "default": "right",
               "enum": ["right", "left"]
             },
+            "animation_type": {
+              "type": "string",
+							"default": "slide_down",
+							"description": "Animation type for menu",
+							"enum": ["slide_down", "slide_up", "none"]
+            },
+            "animation_duration":{
+              "type": "integer",
+              "default": 250,
+              "description": "Duration of animation in milliseconds"
+            },
             "actions": {
               "$ref" : "#/widgets/buttons-grid/properties/actions"
             }
--- a/src/configSchema.json
+++ b/src/configSchema.json
@@ -425,9 +425,9 @@
             },
             "animation_type": {
               "type": "string",
-							"default": "slide_down",
-							"description": "Animation type for menu",
-							"enum": ["slide_down", "slide_up", "none"]
+              "default": "slide_down",
+              "description": "Animation type for menu",
+              "enum": ["slide_down", "slide_up", "none"]
             },
             "animation_duration":{
               "type": "integer",
--- a/src/controlCenter/widgets/menubar/menubar.vala
+++ b/src/controlCenter/widgets/menubar/menubar.vala
@@ -17,7 +17,9 @@ namespace SwayNotificationCenter.Widgets {
         string ? label;
         Position ? position;
         Action[] actions;
-        Gtk.Box ? menu;
+        Gtk.Revealer ? revealer;
+        int animation_duration;
+        Gtk.RevealerTransitionType ? animation_type;
     }
 
     public struct Action {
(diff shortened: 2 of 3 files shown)


ErikReider:
I'll look at this tomorrow. Ping me if I forget :)

src/configSchema.json
```diff
@@ -423,6 +423,17 @@
               "default": "right",
               "enum": ["right", "left"]
             },
+            "animation_type": {
+              "type": "string",
+							"default": "slide_down",
+							"description": "Animation type for menu",
+							"enum": ["slide_down", "slide_up", "none"]
```
ErikReider, line 430: Wrong indent type

src/controlCenter/widgets/menubar/menubar.vala
```diff
@@ -160,14 +167,42 @@ namespace SwayNotificationCenter.Widgets {
...
+                string ? animation_type = get_prop<string> (obj, "animation_type");
+                if (animation_type == null) {
+                    animation_type = "slide_down";
+                    info ("No animation-type for menu-object given using default");
+                }
+
+                Gtk.RevealerTransitionType revealer_type;
+
+                switch (animation_type) {
+                    case "none":
```
ErikReider, line 182: Put default here to simplify

src/controlCenter/widgets/menubar/menubar.vala
```diff
@@ -92,14 +94,19 @@ namespace SwayNotificationCenter.Widgets {
 
                     Gtk.Box menu = new Gtk.Box (Gtk.Orientation.VERTICAL, 0);
                     menu.get_style_context ().add_class (obj.name);
```
ErikReider, line 96: Forgot to mention this before, we should avoid setting the class of the menu and the buttons if the `obj.name` is null. My bad 😅

src/controlCenter/widgets/menubar/menubar.vala
```diff
@@ -92,14 +94,19 @@ namespace SwayNotificationCenter.Widgets {
...
                     show_button.clicked.connect (() => {
-                    bool visible = !menu.visible;
+                    bool visible = !r.get_reveal_child ();
                     foreach (var o in menu_objects) {
-                        o.menu ?.set_visible (false);
+                        o.revealer ?.set_reveal_child (false);
                     }
-                    menu.set_visible (visible);
+                    r.set_reveal_child (visible);
                 });
```
ErikReider, line 110: Indent this. This is a really annoying Uncrustify issue that I've not been able to solve yet...

src/controlCenter/widgets/menubar/menubar.vala
```diff
@@ -17,7 +17,9 @@ namespace SwayNotificationCenter.Widgets {
         string ? label;
         Position ? position;
         Action[] actions;
-        Gtk.Box ? menu;
+        Gtk.Revealer ? revealer;
+        int animation_duration;
+        Gtk.RevealerTransitionType ? animation_type;
```
ErikReider, line 22: Remove the question mark. Shouldn't be nullable

ErikReider (changes requested):
LGTM otherwise! :)

ErikReider:
Oh yeah, also forgot to mention that there's a weird graphical issue when opening two menus. Looks like the 2nd menu opens over the first one. I'm not sure if that's due to them expanding/contracting at the same time resulting in that behavior or not. Just letting you know :)

JannisPetschenka:
> Oh yeah, also forgot to mention that there's a weird graphical issue when opening two menus. Looks like the 2nd menu opens over the first one. I'm not sure if that's due to them expanding/contracting at the same time resulting in that behavior or not. Just letting you know :)

Yeah, I'm aware of that issue :) The only way to fix this I can think of is to somehow wait for `animation_duration` ms after closing the boxes before opening the correct one. But that would double the time it takes to open a drop-down :upside_down_face:.. So I thought about just calling it a feature rather than a bug :laughing: But I can look into changing it if you prefer :)

ErikReider (approved):

ErikReider:
Thanks again for your help! :)

PR #235, opened 2023-03-19
Per app volume control

JannisPetschenka:
Haven't tested everything yet, but apart from that should be ready :)


ErikReider:
I'll try to get to this tonight. If not, ping me :)

src/controlCenter/widgets/volume/volume.vala
```diff
@@ -56,5 +129,35 @@ namespace SwayNotificationCenter.Widgets {
...
+                var s = (SinkInputRow) row;
+                if (s.sink_input.cmp (sink)) {
+                    levels_listbox.remove (row);
+                    break;
+                }
+            }
+            if (levels_listbox.get_children ().length () == 0) {
+                print ("EMPTY\n");
+            }
+        }
```
ErikReider, line 161: Seems like this one isn't working. None of the sink-inputs get removed when ex closing and opening Spotify.

image

ErikReider (changes requested):
Sorry for the delay. Finally had some free time after "finishing" blur work in SwayFX :)

JannisPetschenka:
> Sorry for the delay. Finally had some free time after "finishing" blur work in SwayFX :)

No worries :smile: I'm aware of the issue with deleting elements. The reason is that I close the client when the cc is closed. And since the `remove_active_sink` signal is only send while the client is active, the remove function is never called....

To solve this I could either never close the client or always update the active sinks when opening the cc... But solutions seem sub-optimal to me. I'll see if I can find a better way :)

ErikReider:
> or always update the active sinks when opening the cc... But solutions seem sub-optimal to me.

If there's no other way, this would be the more optimal solution :)

JannisPetschenka:
Finally found the time to look into it again :) Should work now :smile: 
Currently if there are no active sink_inputs you can still reveal the levels_listbox and it will just show an empty list... Might change that but apart from that it should be ready :)

src/controlCenter/widgets/volume/pulseSinkInput.vala
```diff
@@ -0,0 +1,110 @@
...
+                client.set_sink_input_volume (sink_input, (float) scale.get_value ());
+                scale.tooltip_text = ((int) scale.get_value ()).to_string ();
+            });
+        }
+
+        public void update (PulseSinkInput sink_input) {
+            this.sink_input = sink_input;
+
+            icon.set_from_icon_name (
+                sink_input.application_icon_name ?? "application-c-executable",
```
ErikReider, line 100: Use `"application-x-executable"` instead

src/controlCenter/widgets/volume/pulseSinkInput.vala
```diff
@@ -0,0 +1,110 @@
...
+        public static uint32 get_hash_map_key (uint32 i) {
+            return i;
+        }
+
+        construct {
+            volume_operations = new LinkedList<Operation> ();
+        }
+    }
+
+    public class SinkInputRow : Gtk.ListBoxRow {
```
ErikReider, line 63: Move class into its own file

ErikReider (changes requested):

JannisPetschenka:
Fixed the issues and added a label to be displayed when the list of active sink inputs is empty :) Should be ready now :smile:

ErikReider:
Probably just a line ending issue

ErikReider (approved):
LGTM! Thanks again for all of the help :)

PR #236, opened 2023-03-23
Use rounding to prevent incorrect backlight percentage calculation

Mirei124:
Suppose the brightness of screen is 15%, correspondingly the value of `/sys/class/backlight/amdgpu_bl0/brightness` is 38, and `max_brightness` is 255, 38/255 = 0.149, but the output of the calc_percent function is 14, causing the screen brightness will be decreased by 1% when open panel every time.


ErikReider:
I'll try to get to this tonight. If not, ping me :)

ErikReider (approved):
LGTM! Just fix the CI linter errors and I'll merge :)

PR #246, opened 2023-04-24
Little rework of the Fedora Silverblue installation instructions

foopsss:
Change explained in the commit.


ErikReider (approved):
Thanks! :)

Issue #245, opened 2023-04-23
[Bug] Backlight & KB brightness sliders dimmed each time open control center


MrPenguin07:
**Describe the bug**

With the backlight and KB brightness modules active, each time I open the control center (either with sway binding/in terminal with 'swaync-client -t -sw', or via waybar tray icon using the provided setup config), my KB / screen brightness is decreased by ~20% until the brightness levels are at 0%.

Actually if I have the backlight/KB modules ordered as;
1. backlight, 2. KB,
The bottom module (KB) will be reduced 20% each panel open until it's at zero, and only when it is at 0 will the next panel open reduce backlight brightness until it also is reduced to zero. 

So effictively if I turn my KB and backlight to 100% brightness, then open/close the panel maybe 8-10 times in quick succession both my KB and backlight are off.

**To Reproduce**
1. Enable backlight/KB modules
2. Open panel with keybinding or waybar icon
3. Observe brightness sliders being turned down

**Expected behavior**
Brightness levels to not be affected by panel being opened.

**Screenshots**
First time opened panel:
![image](https://user-images.githubusercontent.com/127086564/233819569-0c6e6344-ccc3-4332-a56b-5d6bdd7d5ae4.png)

Immediately close it, then reopen:
![image](https://user-images.githubusercontent.com/127086564/233819578-7e42ad33-a983-4f37-ba15-abdf14ab9286.png)

3rd open, notice the KB brightness became zero, so now it moves onto dimming the backlight...
![image](https://user-images.githubusercontent.com/127086564/233819586-7640b802-d798-4cb9-9ad4-e3227282ddde.png)

**Desktop (please complete the following information):**
 - OS: Gentoo
 - Swaync 0.8.0

** config **

"widgets": [
    "label",
    "menubar",
    "buttons-grid",
    "backlight",
    "backlight#KB",
    "volume",
    "mpris",
    "title",
    "dnd",
    "notifications"
  ],
  "widget-config": {
    "title": {
      "text": "Notifications",
      "clear-all-button": true,
      "button-text": "Clear All"
    },
    "backlight": {
      "label": "",
      "device": "acpi_video0",
      "min": 10
    },
    "backlight#KB": {
      "label": "",
      "device": "system76_acpi::kbd_backlight",
      "subsystem": "leds"
    },
    "volume": {
      "label": ""
    },

ErikReider:
Have you tried the latest changes in master?

MrPenguin07:
[...]

PR #248, opened 2023-04-29
Fix issue #245

JannisPetschenka:


ErikReider:
What was the actual cause of the issue? Shouldn't we calculate the percent on LEDs?

JannisPetschenka:
We could, but that might be bad/annoying depending on the laptop/KB-backlight. For instance my laptop only has 4 brightness levels for the KB so converting that to percent and back confusing since the brightness won't change continuously

ErikReider (approved):
LGTM! Thanks for the quick fix ❤️

PR #254, opened 2023-05-07
Add a tip with gtk_debug env var in README.md

Seme4eg:
I know that info can be dug up in the internet, but it also doesn't take much space in the readme, so why not leave it there. Would have helped me if it was there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ See `swaync(5)` man page for more information
 
 To reload the config, you'll need to run `swaync-client --reload-config`
 
-The main CSS style file is located in `/etc/xdg/swaync/style.css`. Copy it over to your `~/.config/swaync/` folder to customize without needing root access.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -177,7 +177,7 @@ See `swaync(5)` man page for more information
 
 To reload the config, you'll need to run `swaync-client --reload-config`
 
-The main CSS style file is located in `/etc/xdg/swaync/style.css`. Copy it over to your `~/.config/swaync/` folder to customize without needing root access.
+The main CSS style file is located in `/etc/xdg/swaync/style.css`. Copy it over to your `~/.config/swaync/` folder to customize without needing root access. **Tip**: running swaync like this `GTK_DEBUG=interactive swaync` will open a inspector window that'll allow you to see all of the CSS classes + other information.
```
ErikReider, line 180: Could you make this two lines?

ErikReider (changes requested):
Good idea! Thanks :)

ErikReider (approved):
LGTM! :)

Issue #255, opened 2023-05-10
[Bug] swaync won't respect order of rules for visibility


diesphink:
**Describe the bug**

swaync won't respect order of rules for visibility: at loading time the rules get in random (unexpected, but consisting) order.
The order seems to be based on the name used for each rule.

**To Reproduce**

Configure a few notification-visibility rules, but one of them will always run if nothing else executes.

For example, from my cfg:

```json
"notification-visibility": {
  "1-power-status": {
    "state": "transient",
    "app-name": "poweralertd"
  },
  "2-hud-messages": {
    "state": "transient",
    "category": "hud"
  },
  "3-everything-else": {
    "state": "muted",
    "summary": ".*"
  }
},
```

Every notification is getting muted, instead of just the notifications that the first two rules didn't catch.

**Expected behavior**

Notifications either from app-id poweralertd or with category hud should be transient, everything else should be muted.

**Please provide logs if you're experiencing notification errors / bugs**

When I run swaync with debug I can see the order that it's loading from the file, the example above becomes:

```json
"notification-visibility" : {
  "3-everything-else" : {
    "summary" : ".*",
    "state" : "muted"
  },
  "1-power-status" : {
    "app-name" : "poweralertd",
    "state" : "transient"
  },
  "2-hud-messages" : {
    "category" : "hud",
    "state" : "transient"
  }
},
```

As the 3-everything-else is the first one, it explains the behavior.

If I change some of the names (in the case below, added ! at the end of the second rule name), I can randomly change the order:

```json
"notification-visibility" : {
  "2-hud-messages!" : {
    "category" : "hud",
    "state" : "transient"
  },
  "3-everything-else" : {
    "summary" : ".*",
    "state" : "muted"
  },
  "1-power-status" : {
    "app-name" : "poweralertd",
    "state" : "transient"
  }
},
```

The one that eventually worked in the order that I wanted was adding ! to the third rule:

```json
"notification-visibility" : {
  "1-power-status" : {
    "app-name" : "poweralertd",
    "state" : "transient"
  },
  "2-hud-messages" : {
    "category" : "hud",
    "state" : "transient"
  },
  "3-everything-else!" : {
    "summary" : ".*",
    "state" : "muted"
  }
},
```

**Desktop (please complete the following information):**

OS: Ubuntu Sway 23.04 x86_64 
DE: sway (Wayland)
swaync: 0.7.3

**Additional context**
[...]

PR #258, opened 2023-05-11
Preserve config object item order

ErikReider:
Fixes: #255

Preserves the order of which config object items are parsed


diesphink (approved):
Works for me

Issue #62, opened 2022-01-24
[feature request] Bind the client window to the output


nwg-piotr:
Nothing really serious, but it would be not bad to be able to show the client window on a selected output only, e.g. like:

```
        -o, --output    Open the client window on a certain output
```

ErikReider:
@nwg-piotr Finally fixed now when GTk4 is used! :D

PR #262, opened 2023-05-14
Added ability to choose which output to use

ErikReider:
Supersedes: #64 

Fixes: #62 

- [x] Choose monitor as config option
- [x] Notification window logic
- [x] Control center logic
- [ ] Test if disconnecting monitor breaks this
- [ ] Update man page
- [x] Open on focused monitor if the chosen display isn't found

### DIFF
--- a/src/config.json.in
+++ b/src/config.json.in
@@ -4,6 +4,7 @@
   "positionY": "top",
   "layer": "top",
   "layer-shell": true,
+  "output": "auto",
   "cssPriority": "application",
   "control-center-margin-top": 0,
   "control-center-margin-bottom": 0,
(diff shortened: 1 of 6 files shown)


ErikReider:
Need to investigate segfault. Steps to reproduce:

1. Set output to second monitor
2. open CC
3. disable/unplug output
4. CC + blankwindows should disappear (should do this????)
5. Toggle CC (need to also try opening and closing instead of toggling)

ErikReider:
Will try to squeeze this in after the GTK4 port if it's stable enough

nebunebu:
Was there progress made on this?

VAWVAW:
I have rebased this (https://github.com/VAWVAW/SwayNotificationCenter/tree/choose-output) and it seems to work.

The problem with unplugging a monitor while the CC is open still persists but is not specific to this change.

I have also added the ability to refer to the monitors by model. To use the exact format generated by `swaymsg -t get_output` we need to wait for the update to gtk4 as there is no way to get the monitor serial number.
So instead of `Some Company ABC123 0x00000000` it is `Some Company ABC123`.

ErikReider:
Closing in favor of #579

Issue #249, opened 2023-05-04
Allow for positionY: center


Seme4eg:
Currently i found it impossible, but maybe there's a hacky way? To make floating notifications appear for example on right-center

ErikReider:
Not currently possible, would need a small/simple PR

PR #264, opened 2023-05-16
Update swaync.5.scd #249

Seme4eg:
add positionY center


ErikReider:
Oh yeah! I totally for got to add the new config options... Could you include them here? :)

The description should match the one in config_schema.json so should just be a copy/paste job :)

PR #283, opened 2023-06-14
Update i3status-rs Example format syntax for 0.31.0

wseaton:
The formatting syntax has changed slightly, ref: https://greshake.github.io/i3status-rust/i3status_rs/blocks/notify/index.html#examples


ErikReider (approved):
Thanks :)

PR #286, opened 2023-06-26
Added "relative-timestamps" config to switch between "25 minutes ago" and "2023-06-25T20:05:27-07"

dev-dev-dev-dev-dev-dev-dev-dev:
*relative-timestamps*
	type: bool
	default: true
	description: Display notification timestamps relative to now e.g. \"26 minutes ago\". If false, a local iso8601-formatted absolute timestamp is displayed. 

If you think I'm going down the right direction with this, happy to expand to a user-configurable time format feature. I'm personally not a fan of relative time, instead I like to see the exact time a notification was generated for certain items.

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -311,6 +311,8 @@ namespace SwayNotificationCenter {
         public static void reload_config (ModifyNode modify_cb = () => {}) {
             // Re-check if config file path still exists
             string path = Functions.get_config_path (_path);
+            path = File.new_for_path (path).get_path () ?? path;
+            message ("Loading Config: \"%s\"", path);
 
             ConfigModel m = null;
             try {
@@ -331,7 +333,7 @@ namespace SwayNotificationCenter {
                 }
                 m = model;
             } catch (Error e) {
-                stderr.printf (e.message + "\n");
+                critical (e.message);
             }
             _instance = m ?? new ConfigModel ();
             _path = path;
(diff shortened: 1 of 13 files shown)


src/notification/notification.vala
```diff
@@ -556,10 +556,15 @@ namespace SwayNotificationCenter {
         }
 
         public void set_time () {
-            this.time.set_text (get_readable_time ());
+            if(ConfigModel.instance.relative_timestamps) {
+                this.time.set_text (get_relative_time ());
+            }
+            else {
```
ErikReider, line 562: Small nit: else on the same line as `{` as well as the linting issues

src/configModel/configModel.vala
```diff
@@ -521,6 +521,10 @@ namespace SwayNotificationCenter {
         /** Whether to expand the notification center across both edges of the screen */
         public bool fit_to_screen { get; set; default = true; }
 
+        /** Display notification timestamp relative to now e.g. "26 minutes ago".
+        If false, a local iso8601-formatted absolute timestamp is displayed. */
```
ErikReider, line 525: Nit: Should look like this:

```vala
/**
 * Display notification timestamp relative to now e.g. "26 minutes ago".
 * If false, a local iso8601-formatted absolute timestamp is displayed.
 */
```

ErikReider (changes requested):

ErikReider (approved):
LGTM! Thanks for your patience :)

ErikReider:
Fix the trailing whitespace and I'll merge this :)

Issue #295, opened 2023-07-07
Themes not applying


ErikReider:
### Discussed in https://github.com/ErikReider/SwayNotificationCenter/discussions/294

<div type='discussions-op-text'>

<sup>Originally posted by **Boh132Boh** July  6, 2023</sup>
I copied [these dotfiles](https://github.com/lvntcnylmz/dotfiles/tree/main/.config/swaync)  in /home/user/.config/swaync/... and used swaync-client --reload-config, but the notification center didn't change. So I tried to stop and restart swaync, but nothing changed. Then I tried restarting the computer, recompiling and reinstalling following [Debian's guide for backporting](https://wiki.debian.org/SimpleBackportCreation) and tried reloading the css and cofig files in any way I knew and was suggested, but I couldn't manage to get the theme applied.

For context I'm using Debian 12 "Bookworm" and compiled the latest version on my own since I wish to use features like hte volume and brightness control and the version in the repos doesn't support those. 

I already got some [help on reddit](https://www.reddit.com/r/swaywm/comments/14qsn8s/themes_not_applying_in_sway_notification_center/), but we couldn't find out how to solve. 

Attached a screen recording:
https://github.com/ErikReider/SwayNotificationCenter/assets/82704295/5019352d-9785-4068-84e0-eab3f856ed61

Any idea on how to solve? Thank you in advance!</div>

PR #296, opened 2023-07-07
Fixed Debian packaged configs being checked before user configs

ErikReider:
Fixes: #295, #294

### DIFF
--- a/src/functions.vala
+++ b/src/functions.vala
@@ -106,10 +106,7 @@ namespace SwayNotificationCenter {
 
         public static string get_style_path (owned string ? custom_path,
                                              bool only_system = false) {
-            string[] paths = {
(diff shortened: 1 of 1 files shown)


ErikReider:
@Boh132Boh, does this fix the issue?

Boh132Boh:
> @Boh132Boh, does this fix the issue?

When built and run with `ninja -C build && ./build/src/swaync` it works now, when as a debian package still doesn't work.

ErikReider:
> When built and run with `ninja -C build && ./build/src/swaync` it works now, when as a debian package still doesn't work.

The package maintainer would have to include this patch in a new package revision until I release a new version. I have very little time atm so may be a while...

In the mean time, you could just execute the swaync that you built in your sway config file instead :)

Issue #271, opened 2023-05-29
[Feature] Make pulseaudio optional


freijon:
With 0.8.0 some pulseaudio features were added, which is great! At least for most people... I'm packaging this application for the Gentoo Linux Distribution. As you may know, Gentoo allows the user to enable/disable certain features at compile time. We made the pulseaudio stuff an optional feature because of the (quite heavy) pulseaudio dependency and we'd like to offer this application also to users who decided to use a different sound server.

That's why I'm asking to have an option to enable/disable the pulseaudio feature via the `meson_options.txt`. Since 0.8.0 we have to maintain a quite substantial patch to remove everything pulseaudio related. Having this as an optional feature would be very helpful :)

I'm interested to learn what you think about this idea.

ErikReider:
That sounds like a reasonable request :)

How high of a priority is this?

freijon:
I see you added this to the v1.0.0 milestone 👍 Depends how many releases you plan until then. We're all set now for v0.9.0, so no issue there. But for every release I have to figure out what to patch out. This gets more complex the more the pulseaudio code base grows. It takes some time and I might miss something. Of course I always check if the patched code compiles, but there might be other stuff, e.g. in the man pages or in the help that I miss. So I guess the sooner the better from my point of view ;)

PR #297, opened 2023-07-08
Add meson_option to build without PulseAudio Widget

ErikReider:
Fixes: #271

`meson build -Dpulse-audio=false` configures it to build without PulseAudio

### DIFF
--- a/man/swaync.5.scd
+++ b/man/swaync.5.scd
@@ -224,8 +224,10 @@ config file to be able to detect config errors
 			optional: true ++
 		*buttons-grid*++
 			optional: true ++
+#START pulse-audio
 		*volume*++
 			optional: true ++
+#END pulse-audio
 		*backlight*++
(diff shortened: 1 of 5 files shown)


ErikReider:
@freijon If this works as expected on your end, I'll merge it :)

freijon:
Hello @ErikReider 

Thanks! I created a test package and the new option works like a charm.
The only thing I notice is that the man-pages still have the the pulse-stuff in them even if built without pulse. I patched that out as well in our package. I'd consider this a bonus goal though ;)

ErikReider:
I'll check if there's a way of patching out sections through meson?

ErikReider:
> The only thing I notice is that the man-pages still have the the pulse-stuff in them even if built without pulse

Added a "hacky" way of doing this. It works like a charm though :)

Issue #300, opened 2023-08-03
[Feature] Label for per-app volume sliders


pinpox:
**Please read through the README and the Man pages before submitting**
**Please also make sure that there isn't any prior issue describing this feature**

**Is your feature request related to a problem? Please describe.**

The per-app volume sliders are great, but they have now description of what slider controls what application. See this screenshot as an example:
![image](https://github.com/ErikReider/SwayNotificationCenter/assets/1719781/972fd33a-430d-4bf4-b684-604230940330)
This makes them a bit useless as you have to guess what slider does what. I

**Describe the solution you'd like**
t would be great to be able to show the app name next to them, like it is shown in pavucontrol for example.
![image](https://github.com/ErikReider/SwayNotificationCenter/assets/1719781/95986813-6380-44a3-837f-947e4c222292)

**Describe alternatives you've considered**
I have looked at `swaync(5)` couldn't find any setting to show that information and have to assume it is not yet supported. Please let me know if I missed something.

**Additional context**
Have a nice day!

JannisPetschenka:
Weird, for me it displays the actual icon for the application (at least for most of them).. I'll look into it, displaying the app-name should be an easy fix :)

JannisPetschenka:
![20230807_18h07m02s_grim](https://github.com/ErikReider/SwayNotificationCenter/assets/78504175/8a7281ca-4d22-4d1f-bd71-213276bcb9f9)

This is what it would look like with #301

pinpox:
> This is what it would look like with #301

Very nice! When will that be released?

ErikReider:
> > This is what it would look like with #301
> 
> Very nice! When will that be released?

I'll be merging this PR shortly so it'll be available in the master branch pretty soon :)

I probably won't be releasing an official release anytime soon unless there's a big feature that gets merged like the GTK4 port

PR #301, opened 2023-08-07
Add show-per-app-label for volume

JannisPetschenka:
Relating to issue #300

### DIFF
--- a/src/controlCenter/widgets/volume/sinkInputRow.vala
+++ b/src/controlCenter/widgets/volume/sinkInputRow.vala
@@ -3,14 +3,18 @@ namespace SwayNotificationCenter.Widgets {
 
         Gtk.Box container;
         Gtk.Image icon = new Gtk.Image ();
+        Gtk.Label label = new Gtk.Label ("");
         Gtk.Scale scale = new Gtk.Scale.with_range (Gtk.Orientation.HORIZONTAL, 0, 100, 1);
 
         public unowned PulseSinkInput sink_input;
 
         private unowned PulseDaemon client;
 
-        public SinkInputRow (PulseSinkInput sink_input, PulseDaemon client, int icon_size) {
+        private bool show_per_app_label;
+
+        public SinkInputRow (PulseSinkInput sink_input, PulseDaemon client, int icon_size, bool show_per_app_label) {
             this.client = client;
+            this.show_per_app_label = show_per_app_label;
 
             update (sink_input);
 
@@ -20,7 +24,11 @@ namespace SwayNotificationCenter.Widgets {
 
             container = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 0);
 
-            container.add (icon);
+            if (show_per_app_label) {
+                container.add (label);
+            } else {
+                container.add (icon);
+            }
 
             container.pack_start (scale);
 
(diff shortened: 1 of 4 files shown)


src/controlCenter/widgets/volume/sinkInputRow.vala
```diff
@@ -20,7 +24,11 @@ namespace SwayNotificationCenter.Widgets {
 
             container = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 0);
 
-            container.add (icon);
+            if (show_per_app_label) {
+                container.add (label);
+            } else {
+                container.add (icon);
+            }
```
ErikReider, line 31: What's your opinion on showing both icon and label? Would that be too much?

src/controlCenter/widgets/volume/sinkInputRow.vala
```diff
@@ -3,14 +3,18 @@ namespace SwayNotificationCenter.Widgets {
 
         Gtk.Box container;
         Gtk.Image icon = new Gtk.Image ();
+        Gtk.Label label = new Gtk.Label ("");
```
ErikReider, line 6: Replace the empty string with `null`

src/controlCenter/widgets/volume/volume.vala
```diff
@@ -51,6 +52,7 @@ namespace SwayNotificationCenter.Widgets {
                 label_widget.set_label (label ?? "Volume");
 
                 show_per_app = get_prop<bool> (config, "show-per-app") ? true : false;
+                show_per_app_label = get_prop<bool> (config, "show-per-app-label") ? true : false;
```
ErikReider, line 55: We should change these to check if the property is actually found:
```vala
bool show_per_app_found;
bool ? show_per_app = get_prop<bool> (config, "show-per-app", out show_per_app_found);
if (show_per_app_found) this.show_per_app = show_per_app;
bool show_per_app_label_found;
bool ? show_per_app_label = get_prop<bool> (
    config, "show-per-app-label", out show_per_app_label_found);
if (show_per_app_label_found) this.show_per_app_label = show_per_app_label;
```

ErikReider (changes requested):
Just a few minor comments. Thanks again Jannis! Real MVP 💪

src/controlCenter/widgets/volume/sinkInputRow.vala
```diff
@@ -3,14 +3,18 @@ namespace SwayNotificationCenter.Widgets {
...
         Gtk.Scale scale = new Gtk.Scale.with_range (Gtk.Orientation.HORIZONTAL, 0, 100, 1);
 
         public unowned PulseSinkInput sink_input;
 
         private unowned PulseDaemon client;
 
-        public SinkInputRow (PulseSinkInput sink_input, PulseDaemon client, int icon_size) {
+        private bool show_per_app_label;
+
+        public SinkInputRow (PulseSinkInput sink_input, PulseDaemon client, int icon_size, bool show_per_app_label) {
```
ErikReider, line 15: Also, this line is a bit long. Mind making it take up multiple lines instead? :)

src/controlCenter/widgets/volume/sinkInputRow.vala
JannisPetschenka: Personally I think it would get too cluttered.. But why not offer both :)  I added the option to enable/disable the label/icon separately :)

src/controlCenter/widgets/volume/sinkInputRow.vala
```diff
@@ -60,3 +65,5 @@ namespace SwayNotificationCenter.Widgets {
         }
     }
 }
+
+
```
ErikReider, line 69: Nit: remove these extra lines, thanks Uncrustify...

ErikReider (changes requested):

src/controlCenter/widgets/volume/sinkInputRow.vala
JannisPetschenka: Mb, switched code editor and the formatter isn't working properly :upside_down_face: should be fine now :)

ErikReider (approved):
Thanks again! :)

Issue #302, opened 2023-08-08
[Bug] Keyboard navigation in panel has strange behavior


pwnfooj716:
Currently, using the arrow keys to go up and down in the panel causes the selected item to skip by two on each key press. Alt+arrow keys produces the expected behavior of skipping by one on each key press which is super unintuitive. If anything, the behavior should be flipped. In addition, Page Up/Down also seems to skip by one on each key press instead of skipping by one screen-full (not sure if that's the intended behavior because it is not documented anywhere). Further, Page Up seems to have a bug where a notification is skipped past if it does not have an icon associated with it***. It's also worth nothing that it is not documented anywhere that Alt+arrow keys and Page Up/Down do anything.

On another note, using the keyboard to navigate the panel does not scroll to keep the selection on screen. If you arrow down until the selected item is off screen, it should automatically scroll so that the item is on screen.

***Steps to reproduce bug with Page Up key:
1) Trigger a notification with an icon (e.g. notify-send --icon=dialog-warning Warning "This is a warning")
2) Trigger a notification **without** an icon (e.g. notify-send "Text Alert" "This notification contains only text")
3) Trigger another notification with an icon (e.g. notify-send --icon=dialog-information Info "This is some important information")
4) Open swaync panel and arrow down to select least recent notification
5) Press PageUp key

Expected behavior: 2nd notification is selected
Actual behavior: Most recent notification is selected

Swaync may not be a keyboard-centric notification daemon like dunst, mako, etc., but what made this project stand out to me versus those others is the fact that it can be controlled both with mouse and keyboard quite nicely and look good while doing so. If these quirks can be ironed out, swaync would definitely be one of he best wayland notification daemon out there.

pwnfooj716:
Wow, thanks so much for the speedy fix! Just tested it and it works. However, the issue for scrolling the panel when the selected notification goes off screen still remains. Should I open up a new issue for this specifically?

PR #305, opened 2023-08-15
Fixed CC list navigation skipping a notification

ErikReider:
Fixes: #302


pwnfooj716 (approved):
Tested changes on my system and everything is working great. Arrow keys work as expected moving up or down by one entry.

ErikReider:
It still won't scroll when navigating with the arrow keys. Will try to find some time to investigate this Sunday :)

PR #329, opened 2023-10-15
flush stdout after subscription events

capezotte:
Otherwise, trying to pipe its output into another command (say, `jq --unbuffered`) will make it buffer up events instead of feeding information in realtime.

Bonus: remove second printf.


ErikReider (approved):
LGTM! Thanks and sorry for the delay :)

PR #334, opened 2023-11-11
Add notification hint to bypass dnd/inhibition

notpeelz:
I use a couple scripts that trigger custom transient notifications using notify-send (e.g cycling through MPRIS players using playerctld). When I enable dnd, it unfortunately hides those notifications as well :(
I tried using `notify-send --urgency critical`, but critical notifications never expire (it overrides `--expire-time`). I tried closing the notification via dbus after a `sleep`, but that gets very messy very quickly.

Instead, I opted to add a hint to bypass swaync's dnd/inhibit mechanisms. It can be used like this:
```sh
notify-send \
  --expire-time 1000 \
  --transient \
  --hint boolean:SWAYNC_BYPASS_DND:true \
  "hello"
```

Let me know what you think :smiley: 

Oh also, should I add this to the manpage/readme? If so, where?

### DIFF
--- a/src/notiDaemon/notiDaemon.vala
+++ b/src/notiDaemon/notiDaemon.vala
@@ -186,12 +186,20 @@ namespace SwayNotificationCenter {
                 synchronous_ids.set (param.synchronous, id);
             }
 
-            // Only show popup notification if it is ENABLED or TRANSIENT
-            if ((state == NotificationStatusEnum.ENABLED || state == NotificationStatusEnum.TRANSIENT)
-                && !control_center.get_visibility ()
-                // Also check if urgency is Critical and not inhibited and dnd
-                && (param.urgency == UrgencyLevels.CRITICAL
-                    || (!dnd && !swaync_daemon.inhibited && param.urgency != UrgencyLevels.CRITICAL))) {
+            bool show_notification = state == NotificationStatusEnum.ENABLED
+                                     || state == NotificationStatusEnum.TRANSIENT;
+            // Don't show the notification window if the control center is open
+            if (control_center.get_visibility ()) {
+                show_notification = false;
+            }
+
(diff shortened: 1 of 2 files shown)


ErikReider:
> I tried using notify-send --urgency critical, but critical notifications never expire (it overrides --expire-time)

Have you tried changing the `timeout-critical` config option? The default is 0 which disables the timeout.

I like the massive simplification that you made so I'll at least merge that if not also the new param if the config change doesn't work out :)

notpeelz:
Oh I didn't know about `timeout-critical`. I don't have anything that makes use of critical notifications in my setup at the moment, but I can see how it's useful for certain notifications you don't want to miss (e.g battery low warning).

Should I add documentation for this hint, or would you rather do it yourself?

ErikReider:
> Oh I didn't know about `timeout-critical`. I don't have anything that makes use of critical notifications in my setup at the moment, but I can see how it's useful for certain notifications you don't want to miss (e.g battery low warning).
> 
> Should I add documentation for this hint, or would you rather do it yourself?

What I meant was that `SWAYNC_NO_SCRIPT` shouldn't be necessary when notifications can provide the Critical hint

notpeelz:
My previous message wasn't very clear. I meant to say that I see value in keeping the two mechanisms separate. For example, let's say I want critical notifications that linger on screen indefinitely AND have custom notifications bypass dnd while honoring the `expire_timeout` parameter. The use case is probably niche, but at least the custom hint (`SWAYNC_BYPASS_DND`) helps decouple the two mechanisms.

ErikReider (approved):
Appreciate the dialog and I agree with you, thanks :)

ErikReider:
Oh wait, the linter is failing, could you fix that real?

notpeelz:
Should be good now :D

Issue #61, opened 2022-01-24
iOS / Cosmic styled notification stacks


ErikReider:
[Cosmic](https://news.itsfoss.com/system76-cosmic-panel/)
![image](https://user-images.githubusercontent.com/35975961/150807140-cf407ab8-c311-46c9-bc34-46bac0e49708.png)

nightly-brew:
I really like it!
Does it allow the user to expand the grouped notifications and interact with each one singularly?

ErikReider:
> I really like it!
> Does it allow the user to expand the grouped notifications and interact with each one singularly?

Yes sir! Clicking on it expands all of the notifications

nightly-brew:
I wonder now: if the notification on top has buttons, are those shown or are they hidden until the group is expanded?

ErikReider:
That screenshot was from a System76 figma document if I'm not mistaken. My guess would be that they are shown after all notifications are expanded

phwelo:
this would be a serious usability improvement 💯

ldelossa:
this is bad ass :+1:

ErikReider:
I have a work in progress locally on my machine with some pretty nice animations :)

I'll post a PR when I've gotten the actual notification logic in place which should be the "easy" part

ErikReider:
https://github.com/ErikReider/SwayNotificationCenter/assets/35975961/c228b177-fb26-4a92-8ace-9b2d14c635c5

Here's a preview ;)

PR #345, opened 2023-11-29
Grouped notifications

ErikReider:
Fixes: #61 

https://github.com/ErikReider/SwayNotificationCenter/assets/35975961/2cf4da93-eff7-4a0b-85e3-cdcea9c4aba3

TODO:
- [x] Keyboard navigation
- [ ] Animate each notification dismiss instead of instantly resizing group to the new size?
- [x] Group by `desktop-entry` if provided
	- [x] Use as title
- [x] Add app icon next to group name?
- [x] Add close all button
- [ ] Gesture to swipe whole group to close all (when collapsed)
- [x] Move urgent groups to the top
- [x] Scroll to top of group when expanding
- [x] Fix hovering collapsed group


rustysys-dev:
Awesome enhancement! I have been waiting for this!! Quick question for your consideration is there anyway to group other than by just desktop file? It might be useful to group by summary as well, otherwise I would have all 250 slack notifications in 1 group.  If it was grouped by summary then we could get the group to distinguish between channels in slack at the very least.

ErikReider:
> Awesome enhancement! I have been waiting for this!! Quick question for your consideration is there anyway to group other than by just desktop file? It might be useful to group by summary as well, otherwise I would have all 250 slack notifications in 1 group. If it was grouped by summary then we could get the group to distinguish between channels in slack at the very least.
> 
> ```
>    string "Slack"
>    uint32 0
>    string ""
>    string "New message in #rustytest"
>    string "rustytest: @Rusty in channel"
>    array [
>       string "default"
>       string "View"
>    ]
>    array [
>       dict entry(
>          string "desktop-entry"
>          variant             string "Slack"
>       )
>       dict entry(
>          string "urgency"
>          variant             byte 1
>       )
>       dict entry(
>          string "sender-pid"
>          variant             int64 1212
>       )
>    ]
>    int32 -1
> ```

That could probably be added in the future. There might be a notification standard for grouped notifications so that would be better

PR #352, opened 2023-12-14
[New Dependency: sassc] Move to SCSS from regular CSS

ErikReider:
This makes it easier to separate the style file into separate files, allows for nested classes, and other neat SCSS features.

This *should* also make it easier for third-parties to make custom themes.

**NOTE**: This PR adds `sassc` as a dependency


maplepy:
Hello not sure why or if you can help with that but whenever i use `style.scss` instead of `style.css`, I get the default config, although I copied yours from the repo

Issue #361, opened 2023-12-29
[Bug] Text color on light GTK theme


1over137:
**Describe the bug**
Text color is unreadable on light theme. Tested with Adwaita and Arc themes, set using nwg-look. First image is light theme.
![20231229-182842](https://github.com/ErikReider/SwayNotificationCenter/assets/22680475/14d62dd9-fd0b-43fc-a9e5-4813db13d87d)
![20231229-182815](https://github.com/ErikReider/SwayNotificationCenter/assets/22680475/aa18e2cf-f398-41fe-8145-6898ac5d3f5f)

In addition, setting the text-color variable in the css does not change the color of the text "Notifications" or "Do not disturb".

**Desktop (please complete the following information):**
 - OS: Gentoo
 - Version: 0.9

PR #362, opened 2023-12-29
Fix text color on light theme

1over137:
#361
This applies to the changes I made to style.css to the default stylesheets, so that it would use @text-color for the color of the title text and control center text. Please test this before merging.


ErikReider (approved):
Thanks a ton! LGTM :)

PR #368, opened 2024-01-06
Fix mpris icon size when using app or default icon

abmantis:
set_from_gicon and set_from_icon_name receive a Gtk.IconSize enum value,
not an int, so we need to call set_pixel_size to set the proper size.


abmantis:
Closing in favor of #370

PR #370, opened 2024-01-06
Fix default mpris icon size

abmantis:
The widget was was using the .ui size instead of the user config
size in the following situations:

- there was no media player app
- the media player app icon was used
- the audio-x-generic-symbolic icon was used

This fix sets the default size for the album art icon according to the
user's config.

set_from_gicon and set_from_icon_name receive a Gtk.IconSize enum value,
not an int, so we need to call set_pixel_size to set the proper size.


src/controlCenter/widgets/mpris/mpris_player.vala
```diff
@@ -380,11 +381,11 @@ namespace SwayNotificationCenter.Widgets.Mpris {
                 icon = desktop_entry.get_icon ();
             }
             if (icon != null) {
-                album_art.set_from_gicon (icon, mpris_config.image_size);
+                album_art.set_from_gicon (icon, Gtk.IconSize.DIALOG);
```
ErikReider, line 384: These should be `Gtk.IconSize.INVALID`

ErikReider (changes requested):
Other than that, LGTM :)

ErikReider:
Could you also update the branch off of main?

src/controlCenter/widgets/mpris/mpris_player.vala
```diff
@@ -96,6 +96,7 @@ namespace SwayNotificationCenter.Widgets.Mpris {
                     update_buttons (source.media_player.metadata);
                 });
             });
+            album_art.set_pixel_size(mpris_config.image_size);
```
ErikReider, line 99: Oh wait, seems like the linter isn't happy about this line. Could you fix that?

ErikReider (changes requested):

src/controlCenter/widgets/mpris/mpris_player.vala
abmantis: Done. There are a lot of linter changes showing up, so I missed this one (I did not want to add unrelated line changes n this PR).
Maybe it would be nice to lint the whole project and do a single commit with that.

Issue #359, opened 2023-12-21
[Feature] Control center visibility class


AmirDahan:
**Is your feature request related to a problem? Please describe.**
It would be nice to have a visible/hidden class for when the notification center is open or visible.

**Describe the solution you'd like**
Addition of a visible/hidden class when opening/closing the notification center.
Addition of a visible/hidden class when a notification shows up (Outside the notification center).

**Additional context**
I want to add animations, So far i've added a slide in animation for notifications (Slide from top of screen to position)
But i also want to do this for the notification center, It is possible it it would only play once since there's no way (as far as i'm aware) to detect if the notification center is open or closed.

PR #371, opened 2024-01-08
Add control center visibility class

AmirDahan:
This addresses https://github.com/ErikReider/SwayNotificationCenter/issues/359

Animation demo: 

https://github.com/ErikReider/SwayNotificationCenter/assets/72354122/f80e42ce-3126-4a6e-b7d7-62e54cff4f4c

CSS used:
```
.control-center {
	margin-right: -400px;
	margin-left: 400px;
}

.open .control-center {
	transition: all 0.25s ease-out;
	margin: 0px;
}
```


ErikReider:
There should be a built-in css selector that does this? Not really sure though

AmirDahan:
There is ``:active``, But i don't think it works for this case as the whole window is hidden instead of just some widget.
I've tried doing something like this before but haven't gotten it working before.

ErikReider (approved):
LGTM! Thanks :)

PR #372, opened 2024-01-09
Fix brightness slider choppiness

AmirDahan:
Make setting the brightness function async to prevent stutter on some devices.

This probably doesn't happen on a lot of devices but i found it weird that it happens on one of my devices,
This pr should help resolve it.


src/controlCenter/widgets/backlight/backlightUtil.vala
```diff
@@ -77,7 +77,7 @@ namespace SwayNotificationCenter.Widgets {
             if (monitor != null) monitor.cancel ();
         }
 
-        public void set_brightness (float percent) {
+        public async void set_brightness (float percent) {
```
ErikReider, line 80: You have to change the call to the methods to include a `.begin` (`implicit .begin is deprecated`)

ErikReider (changes requested):

AmirDahan:
That should do it

AmirDahan:
Any updates on this?

PR #389, opened 2024-02-25
Allow running binaries without installing schemas

abmantis:
This allows using `meson devenv` to run the binaries without having to
install the schemas or manually setup XDG_DATA_PATH, which is usefull
during development.

`gnome.compile_schemas()` does not affect the installation, only `meson devenv`.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -4,6 +4,8 @@ install_data('org.erikreider.swaync.gschema.xml',
 
 subdir('icons')
 
+gnome.compile_schemas()
+
 compile_schemas = find_program('glib-compile-schemas', required: false)
 if compile_schemas.found()
   test('Validate schema file', compile_schemas,


data/meson.build
```diff
@@ -4,6 +4,8 @@ install_data('org.erikreider.swaync.gschema.xml',
 
 subdir('icons')
 
+gnome.compile_schemas()
```
ErikReider, line 7: Doesn't seem to update unless I remove the data dir in the compilation dir. Only works the first time I change the schema

Otherwise LGTM :)

ErikReider (changes requested):

data/meson.build
abmantis: That is true. I don't really know how to solve that, but I still think this is useful for development on a system without swaync installed (or on nix, where the schemas are isolated).

ErikReider: Might require something like this: https://mesonbuild.com/Reference-manual_functions.html#custom_target

PR #406, opened 2024-03-24
block signal handlers in toggle button when updating state

### LINKED ISSUE #405: [Bug] Toggle button calls toggle action when updating state
**Please read through the README and the Man pages before submitting**
**Please also make sure that there isn't any prior issue describing this bug**

**Describe the bug**
When the state of a toggle button is changed by the update command, the toggle action is called.

**To Reproduce**
Let's just imagine a WiFi toggle button, steps to reproduce the behavior:
1. Change the WiFi state using a command like 'rfkill' while the Control Center (CC) is closed.
2. Open CC
3. The CC attempts to synchronize the button state with the WiFi state.
4. Consequently, the CC executes the toggle action on the button as if the user had clicked it

**Expected behavior**
The expected behavior is for the toggle action to be blocked when updating the button state.

**Please provide logs if you're experiencing notification errors / bugs**
Logs are not relevant in this scenario.

**Screenshots**

**Desktop (please complete the following information):**
 Doesn't matter

**Additional context**
Based on my understanding of GTK, the toggle signal of the button should be blocked during state changes.

theRealCarneiro:
Properly blocking toggle signal when changing toggle button's state, to avoid calling the connected function 
Fixes #405

### DIFF
--- a/src/controlCenter/widgets/shared/toggleButton.vala
+++ b/src/controlCenter/widgets/shared/toggleButton.vala
@@ -32,13 +32,13 @@ namespace SwayNotificationCenter.Widgets {
               // remove trailing whitespaces
               Regex regex = new Regex ("\\s+$");
               string res = regex.replace (msg, msg.length, 0, "");
-			  GLib.SignalHandler.block(this, this.handler_id);
+              GLib.SignalHandler.block (this, this.handler_id);
               if (res.up () == "TRUE") {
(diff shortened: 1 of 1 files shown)


ErikReider (changes requested):
LGTM other than the linting complaints

ErikReider (approved):
Thanks! :)

PR #415, opened 2024-04-09
Allows Variable Notification Window Height

modestbadger1773:
Allows Variable Notification Window Height

Solves:

@/issues/174
@/issues/274
@/issues/177

swayncwindowheight

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -406,6 +406,12 @@ namespace SwayNotificationCenter {
          */
         public int notification_window_width { get; set; default = 500; }
 
+
+        /**
+         * Notification window's height, in pixels.
+         */
+        public int notification_window_height { get; set; default = 800; }
+
         /** Hides the control center after clearing all notifications */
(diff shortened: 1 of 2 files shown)


ErikReider:
The issue with this is that the layer could extend past the outputs bounds if the height is very large, hiding notifications. That's why I haven't really looked into this...

One solution would be to always have it set to the outputs height while only allowing input inside the used area (setting the surfaces input region). Not sure if this is possible in GTK though

modestbadger1773:
I mean, the same is true for `notification-widow-width`.

I think both should be available to the user with comments next to them to warn of their misuse.

https://github.com/ErikReider/SwayNotificationCenter/assets/165446573/8f1912d1-a5ba-42ee-902b-5ebcb2d291e2

ZarredFelicite:
This would be great!

ErikReider:
I'm working on adding this but where the max height is restricted to the actual layer height, with new animations! :D

Stay tuned...

ErikReider:
Closing in favor of #561. Thanks for the help! :D

PR #420, opened 2024-04-14
Fix invalid data types in `configSchema.json`

apostroll:
Fix two instances inside `configSchema.json` where the invalid data type `bool` is used, instead of `boolean`.


ErikReider (approved):
Thanks! :)

PR #426, opened 2024-05-13
add settings to center buttons-grid

rtgiskard:
buttonsGrid is using flowbox as the container, with default halign set
to FILL, for which users have to tweak the button size to make it really
fill the space or it will not be centered.

This adds options to make the buttonGrid centered properly

### DIFF
--- a/src/controlCenter/widgets/buttonsGrid/buttonsGrid.vala
+++ b/src/controlCenter/widgets/buttonsGrid/buttonsGrid.vala
@@ -15,26 +15,27 @@ namespace SwayNotificationCenter.Widgets {
         public ButtonsGrid (string suffix, SwayncDaemon swaync_daemon, NotiDaemon noti_daemon) {
             base (suffix, swaync_daemon, noti_daemon);
 
-            int column_min = 4;
-            int column_max = 7;
+            Gtk.FlowBox container = new Gtk.FlowBox ();
+            container.set_selection_mode (Gtk.SelectionMode.NONE);
+            pack_start (container, true, true, 0);
 
             Json.Object ? config = get_config (this);
             if (config != null) {
                 Json.Array a = get_prop_array (config, "actions");
                 if (a != null) actions = parse_actions (a);
 
-                int? col_min = get_prop<int> (config, "column-min");
-                if (col_min != null) column_min = col_min;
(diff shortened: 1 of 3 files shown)


src/controlCenter/widgets/buttonsGrid/buttonsGrid.vala
```diff
@@ -15,15 +15,27 @@ namespace SwayNotificationCenter.Widgets {
...
+                if (center != null && center)
+                    container.set_halign(Gtk.Align.CENTER);
+
+                int? col_min = get_prop<int> (config, "column-min");
+                if (col_min != null && col_min > 0)
+                    container.set_min_children_per_line(col_min);
+
+                int? col_max = get_prop<int> (config, "column-max");
+                if (col_max != null && col_max > 0)
+                    container.set_max_children_per_line(col_max);
```
ErikReider, line 37: The linter requires spaces between the function name and the parenthesis. Also, I'd prefer if you use brackets on small if-statments.

Not sure why GitHub isn't allowing me to run my GitHub Linting Actions…

ErikReider (changes requested):
Would be great if the column min/max wouldn't be needed, but that's sadly just how GTK works...

Not sure how one would fix this without needing those config properties. Any ideas?

rtgiskard:
> Would be great if the column min/max wouldn't be needed, but that's sadly just how GTK works...
> 
> Not sure how one would fix this without needing those config properties. Any ideas?

Anyway, 

> Would be great if the column min/max wouldn't be needed, but that's sadly just how GTK works...
> 
> Not sure how one would fix this without needing those config properties. Any ideas?

Maybe use some other container other than flowbox. Anyway, for flowbox, align to center seems to require `set_min_children_per_line()` to be set properly.

src/controlCenter/widgets/buttonsGrid/buttonsGrid.vala
rtgiskard: The vala dev settings on my laptop not seems to be stable enough, reformat with uncrustify, feel free to adjust the code on merge if you like ;)

ErikReider:
For future reference: Here's something that would work as a replacement https://gitlab.gnome.org/GNOME/libadwaita/-/merge_requests/219

PR #476, opened 2024-09-07
backlightUtil: attempt to write to sysfs directly before trying dbus

earboxer:


ErikReider:
Oh I totally missed this, sorry... Why is the DBus implementation not enough?

PR #485, opened 2024-10-18
Add a way to change No Notifications text

### LINKED ISSUE #450: [Feature] Custom "No Notifications" text
Hi, maybe it can be done and I haven't found how, but it would be nice to be able to write any text to the "No Notifications" section. This would help when localizing the text in the notification center. If there is such an option, please help me to customize the "No notifications" text. If such a feature is available, then sorry for the unnecessary request. Thanks in advance for reply. Wish you all the best!
изображение

S0nter:
Now "No Notifications" text can be changed with `"text-empty"` json config

Fixes #450

### DIFF
--- a/src/controlCenter/controlCenter.vala
+++ b/src/controlCenter/controlCenter.vala
@@ -4,6 +4,8 @@ namespace SwayNotificationCenter {
         [GtkChild]
         unowned Gtk.Box notifications_box;
         [GtkChild]
+        unowned Gtk.Label text_empty_label;
+        [GtkChild]
         unowned Gtk.Stack stack;
         [GtkChild]
         unowned Gtk.ScrolledWindow scrolled_window;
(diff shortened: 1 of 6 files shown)


src/controlCenter/controlCenter.vala
```diff
@@ -61,11 +63,13 @@ namespace SwayNotificationCenter {
...
             viewport.add (list_box);
 
+            text_empty_label.set_text (ConfigModel.instance.text_empty);
+
             if (swaync_daemon.use_layer_shell) {
                 if (!GtkLayerShell.is_supported ()) {
                     stderr.printf ("GTKLAYERSHELL IS NOT SUPPORTED!\n");
                     stderr.printf ("Swaync only works on Wayland!\n");
-                    stderr.printf ("If running waylans session, try running:\n");
+                    stderr.printf ("If running wayland session, try running:\n");
```
ErikReider, line 72: Good catch! :)

ErikReider (approved):
Thanks! :D

PR #494, opened 2024-12-19
Fix json parsing error in config.json

zjeffer:
When launching swaync:
```
** (swaync:867354): CRITICAL **: 17:12:29.374: configModel.vala:310: /etc/xdg/swaync/config.json:77:24: Parse error: unexpected number `12', expected character `,'
```
This PR fixes this error.


ErikReider (approved):
Good catch! Thanks! :D

PR #512, opened 2025-02-22
fix: no convert upper case for the first character

rtgiskard:
If the app name is not ASCII, it could add extra invalid bytes to the name string

### DIFF
--- a/src/notiModel/notiModel.vala
+++ b/src/notiModel/notiModel.vala
@@ -168,7 +168,7 @@ namespace SwayNotificationCenter {
             // Set name_id
             this.name_id = this.desktop_entry ?? this.app_name ?? "";
 
-            // Set display_name and make the first letter upper case
+            // Set display_name
             string ? display_name = this.desktop_entry ?? this.app_name;
             if (desktop_app_info != null) {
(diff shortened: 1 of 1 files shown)


src/notiModel/notiModel.vala
```diff
@@ -168,15 +168,15 @@ namespace SwayNotificationCenter {
...
+            // Set display_name
             string ? display_name = this.desktop_entry ?? this.app_name;
             if (desktop_app_info != null) {
                 display_name = desktop_app_info.get_display_name ();
             }
             if (display_name == null || display_name.length == 0) {
                 display_name = "Unknown";
             }
-            this.display_name = display_name.splice (0, 1, display_name.up (1));
+            this.display_name = display_name;
```
ErikReider, line 179: There's an `.is_ascii ()` function that you can use. So just wrap the previous logic into an if-statement, bring back the comment, and I'll merge :)

ErikReider (changes requested):

src/notiModel/notiModel.vala
rtgiskard: It might be better to keep the name as it is

ErikReider (approved):

PR #545, opened 2025-05-03
Add loop option to mpris.

aereaux:
This adds a loop option to mpris so that you you can go past the number of items in the carousel.

Let me know if I did anything wrong adding it to the config or documentation.

Thanks!

PR #570, opened 2025-07-07
Add loop-carousel option to mpris.

aereaux:
Followup from: https://github.com/ErikReider/SwayNotificationCenter/pull/545

This adds a loop option to mpris so that you you can go past the number of items in the carousel.

Let me know if I did anything wrong adding it to the config or documentation.

Thanks!


ErikReider (approved):
Thanks! :D

Issue #522, opened 2025-03-04
[Bug] Can't delete/activate a specific grouped notification with keyboard only


ArijanJ:
**Describe the bug**
There is currently no way to delete a single notification through backspace/enter/arrow keys if they are grouped. Trying to delete it with backspace/del will delete every single notification in the grouping

**To Reproduce**
1. Receive multiple grouped notifications
2. Attempt to delete only one of them

**Expected behavior**
There should be a way to do this without using the mouse, as the non-grouped versions work perfectly and have good UX

**Desktop (please complete the following information):**
 - OS: Mint w/ sway
 - Version: latest git revision as of today

ErikReider:
As a temp fix, you can disable the notification grouping now

Issue #392, opened 2024-02-26
Group notifications using something other than the app (to support browsers)


1ace:
**Is your feature request related to a problem? Please describe.**

Half-way between a bug report and a feature request.

For my use case, grouping by app is also useless, because firefox is the only app I use that sends notifications, so all the notifications from my personal & professional webmails and the three chat systems I have to use on various projects I work on (irccloud, element, discord... I wish I was joking) all get mixed together.

**Describe the solution you'd like**

Ideally there would be a better way to group notifications than by app, so that this feature can become useful (and it really could be useful!)

I'm not quite sure what other meta-data is available to group by; perhaps grouping by icon could work?

**Describe alternatives you've considered**

The request in https://github.com/ErikReider/SwayNotificationCenter/issues/376 (to be able to disable this feature) would remove the problem, but I believe it's also possible to make grouping work.

PR #575, opened 2025-07-09
Make notification grouping optional (#376)

### LINKED ISSUE #376: Grouped notifications should be optional?
Anyway to easily make this a feature we can toggle? I find the new behavior adds alot of mouse movement/clicks.

Thanks for the great software!

ArijanJ:
Notification grouping is cool and looks neat, but it's not very accessible with a keyboard-focused workflow (#522).

This PR adds a "grouping" config entry that defaults to true, but is opt-out as requested in #376.

There is also #392, but that is more of an enhancement request, so this will only close #376.

### DIFF
--- a/man/swaync.5.scd
+++ b/man/swaync.5.scd
@@ -111,6 +111,11 @@ config file to be able to detect config errors
 	default: true ++
 	description: If control center should use keyboard shortcuts
 
+*grouping* ++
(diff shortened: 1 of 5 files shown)


ErikReider (changes requested):
Other than the name, LGTM! :D

PR #576, opened 2025-07-10
feat: new buttons-per-row property for buttons-grid widget

GabMus:
This change adds a new property `buttons-per-grid` to the `buttons-grid` widget configuration.

This was made to avoid having extra space on the right side of the `buttons-grid` widget, caused by not setting the `Gtk.FlowBox` `max-children-per-line` property, which defaults to 7.

This keeps the default of 7 and optionally allows users to override it to better fit this widget to specific use-cases.

Following are screenshots of the result.

Before:

image

After, with value set to 5:

image


ErikReider:
Thanks! :D

PR #577, opened 2025-07-10
Truncate Per App Volume Labels if exceeding a certain (hardcoded, but could also be configurable) length.

Energy0120:
As title says.
It can be paired with the css property min-width while styiling to also adjust shorter names and align volume controllers of any name.

Before:
image

After:
image
(For context i set 100px as the min width to align shorter labels as well)
image

The implementation of max-width and overflow css properties would make this patch not necessary to achieve the same result.

### DIFF
--- a/src/controlCenter/widgets/volume/sinkInputRow.vala
+++ b/src/controlCenter/widgets/volume/sinkInputRow.vala
@@ -56,7 +56,11 @@ namespace SwayNotificationCenter.Widgets {
             }
 
             if (show_per_app_label) {
-                label.set_text (this.sink_input.name);
+                if (this.sink_input.name.length >=10) {
+                    label.set_text(this.sink_input.name.substring(0, 7).concat("...")); 
+                } else {
+                    label.set_text (this.sink_input.name);
+                }
             }
 
             scale.set_value (sink_input.volume);


src/controlCenter/widgets/volume/sinkInputRow.vala
```diff
@@ -56,7 +56,11 @@ namespace SwayNotificationCenter.Widgets {
             }
 
             if (show_per_app_label) {
-                label.set_text (this.sink_input.name);
+                if (this.sink_input.name.length >=10) {
+                    label.set_text(this.sink_input.name.substring(0, 7).concat("...")); 
+                } else {
+                    label.set_text (this.sink_input.name);
+                }
```
ErikReider, line 63: This can be done internally through GTK with:
```vala
label.set_wrap (false);
label.set_single_line_mode (true);
label.set_width_chars (CONFIG_VARIABLE_HERE);
label.set_ellipsize (Pango.EllipsizeMode.END);
```

I'd also like it to be a config option, something like `"show-per-app-label-length": -1` where if the value is below 1, it becomes -1 (tells GTK to calculate the length automatically)

ErikReider (changes requested):

PR #594, opened 2025-07-15
Fixed missing text color on the notification action button

bchmnn:
I noticed that the `--text-color` var is not applied to the notification action buttons. Hence, this PR fixes this issue.

Before:
image

After:
image


ErikReider (approved):
Thanks!

PR #602, opened 2025-07-16
feat: Choose outputs by name

VAWVAW:
#579 added the ability to choose output by connector name. On my system the connector naming isn't always consistent so I would like to refer to monitors by name.

This PR allows to use `"<manufacturer> <model>"` instead of the plug name. This is almost the same output as `swaymsg -t get_outputs` (minus some extra identifier sway shows). Maybe we could also add a `swaync-client --list-outputs` but I'm not sure how to do this.


ErikReider (approved):
Thanks! :)

Issue #344, opened 2023-11-28
[Bug] Missing icons in volume mixer


AmirDahan:
**Please read through the README and the Man pages before submitting**
**Please also make sure that there isn't any prior issue describing this bug**

**Describe the bug**
Icons are missing in the volume mixer

**To Reproduce**
1: Enable show-per-app
2: Enable show-per-app-icon

**Expected behavior**
Icons showcasing which application the sound is coming from to appear next to the sliders

**Screenshots**
![volumemixer](https://github.com/ErikReider/SwayNotificationCenter/assets/72354122/0a64ddab-5820-480a-a7e9-bebc315712ff)

**Desktop (please complete the following information):**
 - OS: Arch linux
 - Version 0.9.0

**Additional context**
I've disabled show per app label since some programs have long names (It would be nice to have some sort of max length or something)

ErikReider:
Is this still an issue?

AmirDahan:
I'm afraid so.
swaync 0.9.0 (git-f68dfa4, branch 'main')

Sohail-XD:
![image](https://github.com/ErikReider/SwayNotificationCenter/assets/86248129/0f836caa-8c10-401e-b720-ebc245f1737b)

The images were working properly a few days ago, I dont know exactly which version but they were working when the "square indicator" on the slider was added. With the latest version they do not get displayed.

Sohail-XD:
I have to apologize but I missed a detail. Mpv, vlc icons are displayed fine. Only firefox icons don't show up.

matthis-k:
It seems Spotify icons also don't show up.
Maybe it hast to do with the icon theme used, or something like that? I remember that being an issue in waybar a while ago and it had something to do with setting an icon theme (I think it was [this module](https://github.com/Alexays/Waybar/wiki/Module:-Taskbar)).

PR #611, opened 2025-07-28
fix: Search icon by process name if icon is missing 

themkoi:
This fixes: #344 
Some apps dont provide the icon so this makes it fallback to searching for the icon by process name


ErikReider (approved):
LGTM! Thanks! :D

PR #620, opened 2025-08-17
Fix `transition` typo

MariaSolOs:
Fixing a typo in these variable names.

~This is a breaking change, and so I'm open to deprecating the older variable name as a more flexible change instead.~ See discussion below for the resolution.


ErikReider (approved):
Thanks again! I should really get spell checking working in vim... 😆

Issue #442, opened 2024-06-17
[Bug] Option needed to hide the 'Settings' button that shows up at the bottom of some notifications


guttermonk:
**Description**
I'm trying to hide the 'Settings' button that shows up on my browser notifications using:  
```
    .notification-action {
      display: none;
    }
```
But this causes the css to error out. I really don't have a need for this button, since I can go into my browser settings at any time to find these settings and I rarely need to change them once they are set. So while the button is convenient, it does take up more space than it is helpful.

**Expected behavior**
I expect the above css code to hide the 'Settings' button.

**Screenshots**
![swappy-2024-06-16-21:20:29](https://github.com/ErikReider/SwayNotificationCenter/assets/4753752/0e1f29d7-a66b-40ee-a25a-436a4e36d1d5)

**Desktop**
 - OS: NixOS 24.05 with Hyprland

ErikReider:
Similar to #466 but the opposite

ErikReider:
This is Chromium adding this button to the notification, so I'd need to add a feature to hide certain actions

guttermonk:
Thanks for the info @Vlad55432. I tried applying the margin and color settings you suggested to .notification-action, but it had too broad of an affect. Any way to use css to target just the settings button from Chrome notifications?

PR #654, opened 2025-10-10
Added Notification action filtering

ErikReider:
Fixes: #442 

Example config for hiding the pesky "Settings" action for Chromium notifications:

```json
  "notification-action-filter": {
    "chromium": {
      "desktop-entry": "chromium-browser",
      "use-regex": false,
      "id-matcher": "settings"
    },
  },
```

Example config for hiding all actions:

```json
  "notification-action-filter": {
    "hide-all-actions": {
      "use-regex": true,
      "id-matcher": ".*"
    }
  },
```

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -202,6 +202,11 @@ namespace SwayNotificationCenter {
                 }
                 node.set_string (eval.value_nick);
                 return node;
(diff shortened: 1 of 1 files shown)


ErikReider:
> /boss $100 thanks

Just installed the Boss app to the repo so try commenting that in the issue thread, thanks :)

PR #712, opened 2026-03-13
Invoke default action when `swaync-client --action` is called without an index

bakhtiyarneyman:
Previously, `--action` without an index would call `click_alt_action(0)`, which silently did nothing if the alt actions box wasn't visible. Now it invokes the notification's default action instead, matching user expectations. When an index is provided, the existing alt action behavior is preserved. Please note, that this is a breaking change: users might need to change their scripts.

### DIFF
--- a/src/client.vala
+++ b/src/client.vala
@@ -80,7 +80,9 @@ private void print_help (string[] args) {
     print ("      \t --close-latest \t\t Closes latest notification\n");
     print ("  -C, \t --close-all \t\t\t Closes all notifications\n");
     print ("  -a, \t --action [ACTION_INDEX]\t " +
-           "Invokes the default action, or action [ACTION_INDEX], of the latest notification\n");
+           "Invokes the action [ACTION_INDEX] (or 0) of the latest notification\n");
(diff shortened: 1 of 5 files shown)


src/client.vala
```diff
@@ -78,7 +80,7 @@ private void print_help (string[] args) {
     print ("      \t --close-latest \t\t Closes latest notification\n");
     print ("  -C, \t --close-all \t\t\t Closes all notifications\n");
     print ("  -a, \t --action [ACTION_INDEX]\t " +
-           "Invokes the action [ACTION_INDEX] of the latest notification\n");
+           "Invokes the default action, or action [ACTION_INDEX], of the latest notification\n");
```
ErikReider: Might be better if an `--action-default` (or something similar) command was added instead. This wouldn't break existing configs.

ErikReider (changes requested):

src/client.vala
bakhtiyarneyman: Done, please take another look.

ErikReider (approved):
Thanks!

PR #713, opened 2026-03-13
Added --close [ID] flag to close a notification by ID

bakhtiyarneyman:


src/client.vala
```diff
@@ -177,6 +180,14 @@ public int command_line (ref string[] args, bool skip_wait) {
             case "-C":
                 cc_daemon.close_all_notifications ();
                 break;
+            case "--close":
```
ErikReider, line 183: `--close` feels a bit ambiguous, maybe `--close-id` or something? Any ideas?

ErikReider (changes requested):
Just a small nit

src/client.vala
bakhtiyarneyman: I have been trying to be consistent with `dunstctl`:
```
...
  action                            Perform the default action, or open the
                                    context menu of the notification at the
                                    given position
  close [ID]                        Close the last notification or the
                                    notification with given ID
  close-all                         Close all the notifications
...
```

`close-id` feels wrong, because you are not closing an id, you are closing a notification. `close-notification` seems long. `close-by-id` is another option. My preference is to just keep `close`. What do you want?

bakhtiyarneyman: By the way, if I add the new mode regarding default action as a new flag, as you requested in a different PR, we'd end up with a weird non-symmetry:

- --close-id 
- --close-all
- --action
- --action-default

As you can see, --action would have no suffix, but --close would.

ErikReider (approved):
Thanks! :D

Issue #277, opened 2023-06-06
[Feature] Calendar Widget


ldelossa:
**Is your feature request related to a problem? Please describe.**
Not really a problem, just a nice to have. 

**Describe the solution you'd like**
GTK ships with a stock calendar widget. 

Could be cool to try to place that as a widget inside SwayNC. 

**Describe alternatives you've considered**
Waybar's hover calendar over the clock widget, but that one kinda sucks.

MusicalArtist12:
I'd also really like to have this as a feature

trevarj:
https://github.com/ErikReider/SwayNotificationCenter/pull/721

Put up a starting PR which pretty much gets me what I want. 
If anyone is interested in taking it over, please feel free.

<img width="400" height="800" alt="Image" src="https://github.com/user-attachments/assets/2aeaa06c-d74d-4373-9cec-5e96d8b75ac0" />

Issue #705, opened 2026-02-23
[Feature] ext-background-effect support


ErikReider:
Support for the background effect Wayland protocol for specifying which regions to explicitly blur.

ryantheerman:
@ErikReider GTK merged in their implementation https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/10145
What do you think? Good enough for use in swaync or better to merge in @trevarj 's pr?

PR #721, opened 2026-04-01
Added calendar widget

trevarj:
The widget is an embedded gtk4 calendar which can be styled as such.

Written with the assistance of Qwen Code.

--- 

This is good enough for me to use on my fork, but I don't have enough Vala or CSS knowledge to go any further.

Going to leave it here in case anyone wants to clean it up or offer some tips.

I think there is probably room for improvement with getting the [GTK Calendar widget CSS ](https://docs.gtk.org/gtk4/class.Calendar.html#css-nodes) to be accessible.

### DIFF
--- a/data/style/widgets/calendar.scss
+++ b/data/style/widgets/calendar.scss
@@ -1,0 +1,34 @@
+:root {
+  --widget-calendar-font-size-day: 1.1rem;
+  --widget-calendar-font-size-date: 1.4rem;
+}
+
+.widget-calendar {
+  padding: 12px;
+  margin: 8px;
+
+  .calendar-container {
+    .day-label {
+      font-size: var(--widget-calendar-font-size-day);
+      font-weight: bold;
+      color: var(--text-color);
+      margin-bottom: 2px;
+    }
+
+    .date-label {
+      font-size: var(--widget-calendar-font-size-date);
+      color: var(--text-color);
+      opacity: 0.8;
+      margin-bottom: 8px;
+    }
+
+    calendar {
+      background: rgba(var(--noti-bg), var(--noti-bg-alpha));
+      border: var(--border);
+      border-radius: var(--border-radius);
+      padding: 8px;
+      color: var(--text-color);
+      font-size: 0.9rem;
+    }
+  }
+}
--- a/src/controlCenter/widgets/calendar/calendar.vala
+++ b/src/controlCenter/widgets/calendar/calendar.vala
@@ -1,0 +1,131 @@
+namespace SwayNotificationCenter.Widgets {
+    public class Calendar : BaseWidget {
+        public override string widget_name {
+            get {
+                return "calendar";
+            }
+        }
+
+        Gtk.Calendar calendar;
+        Gtk.Label date_label;
+        Gtk.Label day_label;
+
+        string day_format = "%A";
+        string date_format = "%B %-d, %Y";
+
+        public Calendar (string suffix) {
+            base (suffix);
+
+            Json.Object ?config = get_config (this);
+
+            var container = new Gtk.Box (Gtk.Orientation.VERTICAL, 6);
+            container.add_css_class ("calendar-container");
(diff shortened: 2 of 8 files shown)


trevarj:
Closes #277 (see screenshot there)

src/controlCenter/widgets/calendar/calendar.vala
```diff
@@ -0,0 +1,72 @@
...
+        public Calendar (string suffix) {
+            base (suffix);
+
+            Json.Object ?config = get_config (this);
+
+            var container = new Gtk.Box (Gtk.Orientation.VERTICAL, 6);
+            container.set_margin_start (12);
+            container.set_margin_end (12);
+            container.set_margin_top (12);
+            container.set_margin_bottom (12);
```
ErikReider: We should set these through CSS so that it's user configurable

src/controlCenter/widgets/calendar/calendar.vala
```diff
@@ -0,0 +1,72 @@
...
+            container.set_hexpand (true);
+
+            day_label = new Gtk.Label ("");
```
ErikReider: Set the string argument to null instead of an empty string

src/controlCenter/widgets/calendar/calendar.vala
```diff
@@ -0,0 +1,72 @@
...
+            day_label.add_css_class ("day-label");
+            day_label.set_halign (Gtk.Align.START);
+
+            date_label = new Gtk.Label ("");
```
ErikReider: Same here

src/controlCenter/widgets/calendar/calendar.vala
```diff
@@ -0,0 +1,72 @@
+using GLib;
```
ErikReider: Not needed

src/controlCenter/widgets/calendar/calendar.vala
```diff
@@ -0,0 +1,72 @@
...
+            container.append (date_label);
+            container.append (calendar);
+            append (container);
+
+            calendar.day_selected.connect (update_labels);
+            update_labels ();
+        }
+
+        void update_labels () {
+            var now = new GLib.DateTime.now_local ();
```
ErikReider: The `GLib.` prefix isn't needed

data/style/widgets/calendar.scss
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  .date-label {
+    font-size: 14px;
+    color: var(--text-color);
+    opacity: 0.8;
+    margin-bottom: 8px;
+  }
+
+  calendar {
```
ErikReider: Might wanna use `.calendar-widget` instead

ErikReider (changes requested):
Just a few nits, looks good overall though :)

What's the use case for this? I don't think that the calendar widget integrates with the systems calendar to display events and such, so is it just for viewing the dates/week number?

Could you also provide a screenshot of how it looks with the day and date labels visible?

trevarj:
> Just a few nits, looks good overall though :)

Thanks for reviewing! 

> What's the use case for this? I don't think that the calendar widget integrates with the systems calendar to display events and such, so is it just for viewing the dates/week number?

As another user wrote on the issue, if you use waybar the calendar widget is pretty bad. This is just a replacement for that and for when you want to quickly look at a date or day.

> Could you also provide a screenshot of how it looks with the day and date labels visible?

I actually went back to a previous version cause I was experiencing a bug on the latest where floating notification backgrounds are randomly not rendering correctly. I will get a screenshot once I solve that and upgrade again to my fork.

trevarj:
Had some time to make the PR feedback changes here and also fix up the SCSS, config options and added the man page doc.

trevarj:
A nice screenshot with #705 

image

ErikReider (approved):
Thanks! :)

Issue #167, opened 2022-09-26
Error when running `swaync`: `COULD NOT FIND CSS FILE! REINSTALL THE PACKAGE!`


perlinm:
If I install `swaync` and try to run it, I simply get the error message `COULD NOT FIND CSS FILE! REINSTALL THE PACKAGE!` and an exit code of 1.  I am running Ubuntu 22.04, and this occurs both if I try to build `swaync` from source and if I try to install it through `nix`.

I'm not sure what other info might help here, but I am willing to answer any questions.

ErikReider:
Run `meson install -C your_build_dir`

perlinm:
I did... Specifically, I ran
```
git clone https://github.com/ErikReider/SwayNotificationCenter.git
cd SwayNotificationCenter
meson build
ninja -C build
meson install -C build
swaync
```
[Here](https://github.com/ErikReider/SwayNotificationCenter/files/9648344/log.txt) is a log that I copy/pasted from my console.

ErikReider:
Oh, my bad... Seems like it's installing those files into `/usr/local` when not finding the `MESON_INSTALL_PREFIX` env variable

ErikReider:
Does #168 fix your issue?

perlinm:
It does, thank you! :slightly_smiling_face: 

[Here's](https://github.com/ErikReider/SwayNotificationCenter/files/9658655/log.txt) a log for posteriority.

PR #727, opened 2026-04-19
Use meson config_path as fallback instead of hardcoded /usr/local

### LINKED ISSUE #726: Fallback config/CSS paths hardcoded to /usr/local break custom prefix installs
**Describe the bug**

When building swaync with a custom prefix (e.g. `--prefix=~/.local`), swaync fails to start because the fallback paths for `config.json` and `style.css` in `src/functions.vala` are hardcoded to `/usr/local/etc/xdg/swaync/`.

This means `meson install` correctly installs the files to `~/.local/etc/xdg/swaync/`, but swaync only searches `/usr/local/etc/xdg/swaync/` as the fallback, so it can't find them.

This affects any installation with a non-default prefix, including:
- User-local installs: `--prefix=~/.local`
- FreeBSD ports: `/usr/local`
- NetBSD pkgsrc: `/usr/pkg`

**To Reproduce**

```sh
meson setup --prefix=~/.local build
ninja -C build
meson install -C build
swaync
```

Output:

```
Starting SwayNotificationCenter version swaync 0.12.6 (git-1043ef9, branch 'main')
COULD NOT FIND CONFIG FILE! REINSTALL THE PACKAGE!
```

After creating `~/.config/swaync/config.json`:

```
COULD NOT FIND CSS FILE! REINSTALL THE PACKAGE!
```

**Expected behavior**

swaync should find the system-level config and CSS files installed under the custom prefix.

**Root cause**

`src/functions.vala` lines 148 and 190 hardcode the fallback path:

```vala
paths += "/usr/local/etc/xdg/swaync/style.css";
paths += "/usr/local/etc/xdg/swaync/config.json";
```

These should use the meson build-time `config_path` instead.

**Related:** #4, #167, #168

ljdi:
Fixes: #726

The hardcoded fallback path `/usr/local/etc/xdg/swaync/` in `src/functions.vala` breaks installations with custom prefixes (e.g. `~/.local`, FreeBSD's `/usr/local`, NetBSD's `/usr/pkg`).

This PR injects the meson `config_path` (resolved at build time) into the Vala source as a compile-time constant, replacing the hardcoded fallback in both `get_style_path()` and `get_config_path()`.

Related: #4, #167, #168

### DIFF
--- a/src/constants.vala.in
+++ b/src/constants.vala.in
@@ -1,5 +1,6 @@
 public class Constants {
     public const string VERSION = @VERSION@;
     public const string VERSIONNUM = @VERSION_NUM@;
+    public const string CONFIGPATH = @CONFIG_PATH@;
     public const uint ANIMATION_DURATION = 400;
 }
--- a/src/functions.vala
+++ b/src/functions.vala
@@ -145,7 +145,8 @@ namespace SwayNotificationCenter {
                                           path, "swaync/style.css");
             }
             // Fallback location. Specified in postinstall.py. Mostly for Debian
-            paths += "/usr/local/etc/xdg/swaync/style.css";
+            paths += Path.build_path (Path.DIR_SEPARATOR.to_string (),
+                                      Constants.CONFIGPATH, "style.css");
 
             info ("Looking for CSS file in these directories:\n\t- %s",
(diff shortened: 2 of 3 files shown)


ErikReider (approved):
Thanks! :)

PR #729, opened 2026-04-22
fix(backlight): recreate FileMonitor after close

### LINKED ISSUE #728: [Bug] Backlight slider stops updating live after control center is closed once
**Describe the bug**
Backlight slider value does not update in real time after the control center has been closed and reopened at least once. First open after swaync start works; every subsequent open, external brightness changes (e.g. `brightnessctl`, XF86MonBrightnessUp/Down) no longer move the slider while the control center is visible. Volume slider works correctly in the same session.

**Root cause**
`src/controlCenter/widgets/backlight/backlightUtil.vala`:

- Constructor creates the `FileMonitor` exactly once.
- `close()` (on CC hide) calls `monitor.cancel()`. GLib `FileMonitor` is single-use; once cancelled it will not emit further `changed` signals.
- `start()` (on CC show) calls `connect_monitor()` which only re-wires the `changed` callback onto the already-cancelled monitor. The monitor is never re-created, so no events fire.

**To Reproduce**
1. Enable backlight widget.
2. Open control center, press XF86MonBrightnessUp — slider updates.
3. Close control center.
4. Open it again, press XF86MonBrightnessUp — slider does not update (brightness itself does change).

**Expected behavior**
Slider updates every time, as the volume slider does.

**Proposed fix**
Recreate the `FileMonitor` in `start()` before `connect_monitor()`, or move monitor creation into `connect_monitor()` itself.

**Desktop**
- OS: CachyOS (Arch)
- swaync: 0.12.6-1.1
- Compositor: Hyprland
- Device: `gmux_backlight` (Apple T2 MacBook)

TimERTy:
## Summary
- `FileMonitor` is single-use: once `monitor.cancel()` runs in `close()` the same instance never emits `changed` again.
- Monitor was created only in the constructor, so after the control center closed the first time, reopening left the slider stale.
- Null `monitor` in `close()` and (re)create it lazily in `connect_monitor()` so every `start()` gets a live monitor.

Fixes #728

### DIFF
--- a/src/configModel/configModel.vala
+++ b/src/configModel/configModel.vala
@@ -347,24 +347,24 @@ namespace SwayNotificationCenter {
             switch (this) {
                 default :
                 return "enabled";
-                case MUTED:
+                case MUTED :
                     return "muted";
-                case IGNORED:
+                case IGNORED :
                     return "ignored";
-                case TRANSIENT:
+                case TRANSIENT :
                     return "transient";
             }
         }
 
         public static NotificationStatusEnum from_value (string value) {
             switch (value) {
-                default:
-                    return ENABLED;
-                case "muted":
+                default :
+                return ENABLED;
+                case "muted" :
                     return MUTED;
-                case "ignored":
+                case "ignored" :
                     return IGNORED;
-                case "transient":
+                case "transient" :
                     return TRANSIENT;
             }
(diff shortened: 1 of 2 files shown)


ErikReider (approved):
Thanks!

PR #732, opened 2026-04-27
Update all `actions/checkout` uses in CI to latest v4

nikosavola:
Update all `actions/checkout` uses in CI to latest v6


ErikReider (approved):
Thanks :)

PR #734, opened 2026-04-27
Cache compiled regexes and guard compilation

nikosavola:
Cache compiled regexes and guard compilation instead of compiling on every call


ErikReider (approved):
Thanks! I moved the notification regex into `static construct` which only gets called once (as the regex variables already are static) :)

PR #735, opened 2026-04-27
Use StringBuilder in `filter_string`

nikosavola:
[`StringBuilder`](https://valadoc.org/glib-2.0/GLib.StringBuilder.html) should have faster appending due to pre-allocation

### DIFF
--- a/src/functions.vala
+++ b/src/functions.vala
@@ -280,14 +280,14 @@ namespace SwayNotificationCenter {
         public delegate bool FilterFunc (char character);
 
         public static string filter_string (string body, FilterFunc func) {
-            string result = "";
+            var result = new StringBuilder ();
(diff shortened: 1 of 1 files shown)


src/functions.vala
```diff
@@ -280,14 +280,14 @@ namespace SwayNotificationCenter {
         public delegate bool FilterFunc (char character);
 
         public static string filter_string (string body, FilterFunc func) {
-            string result = "";
+            var result = new StringBuilder ();
```
ErikReider: If you're worrying about pre-allocation performance, this would be better:
```vala
var result = new StringBuilder.sized (body.length);
```

ErikReider (changes requested):

PR #738, opened 2026-04-27
Replace stderr.printf with structured GLib logging

nikosavola:
Replace stderr.printf with structured GLib logging such as https://docs.gtk.org/glib/func.critical.html


ErikReider:
Thanks for all of your PRs. I'm a bit busy at the moment with my dissertation and such, so might be a while before I can look at the changes

ErikReider (approved):
Thanks yet again! :D

PR #743, opened 2026-05-14
Fix Clear All button leaving phantom rows in control center in some cases

just-barcodes:
play_remove_animation's done signal handler can fire synchronously during adw_animation_play(). Sometimes this means commit_group_removal is skipped and the listbox retains rows that have no visible children.

I tested it on two machines with very similar setups, but can only reliably reproduce it on one of them with the following:

- send some notifications using `notify-send 123` a few times
- click `Clear All` in the control center
- send another notification `notify-send 234` -> it will render below an "invisible gap"

This PR fixes that.

### DIFF
--- a/src/notificationGroup/notificationGroup.vala
+++ b/src/notificationGroup/notificationGroup.vala
@@ -222,7 +222,10 @@ namespace SwayNotificationCenter {
 
             if (get_mapped () && transition) {
                 remove_animation_done_id = remove_animation.done.connect ((e) => {
-                    play_remove_animation.callback ();
+                    Idle.add (() => {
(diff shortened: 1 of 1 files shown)


ErikReider:
Thanks! How often does this bug occur? 

I'm currently a bit busy with writing my dissertation so it'll probably be a few weeks until I have time to review this.

just-barcodes:
On one of my machines it actually happens every time. On another one it never happens.
I attach a short screen recording of the bug:

https://github.com/user-attachments/assets/5c9258e1-3ade-406b-93a6-0f9251648753

ErikReider (approved):
LGTM thanks! :D

PR #748, opened 2026-06-07
Fix loading Mpris album art from URLs without extensions

AlignedTrack432:
Spotify and other streaming services tended to not function, because they often don't include the file extension in the URL, and the previous implementation could not auto-detect it. Here, I've revamped the code for loading album art in order to resolve this issue.

20260607_13h47m01s_grim

It is worth noting that Spotify actually does not work by default. This is due to a well-known, long-standing bug in the Spotify desktop client where it references an older URL for it's album art. However, this fix still applies for Spotify in the browser, along with any other streaming services.


AlignedTrack432:
I'm done, I cannot for the life of me figure out how to properly format this file.

